PR #113, opened 2022-05-30
Add jest and some simple units tests

Robiquet:
I gave unit testing with Cypress a go but it seems quite 'heavy' to use, also doesn't look like it's possible to have the test file inline with the file that's being tested which I think is nice for units tests.

I think the syntax is fairly similar so shouldn't be too much learning overhead. I'll keep this branch as is for now and add more tests if the team is happy with this approach

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,4 @@
+{
+    "presets": ["next/babel"],
+    "plugins": []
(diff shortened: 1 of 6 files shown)


TvrtkoM:
LGTM

Issue #124, opened 2022-06-16
Create both market types with the same extrinsic


Robiquet:
In runtime version 0.3.3 scalar and categorical markets can both be created by the same extrinsic. Previously there were separate extrinsics for each market type. The front end should be updated to reflect these changes

PR #130, opened 2022-06-21
#124 Bump sdk + market creation and pool deployment extrinsic changes

### LINKED ISSUE #131: Market card expansion causes error
After clicking on the market card arrow to expand it I get the following error:

image

yornaath:
Work done:

- [x] Use new SDK createMarket function passing market type as parameters.
- [x] Ensure that amounts of assets are the same amount in accordance to pallet change.
- [x] Set default Authorized address for market dispute mechanism from env var `NEXT_PUBLIC_MDM_AUTHORIZED_DEFAULT_ADDRESS`

Also Fixes #131


pages/create.tsx
```diff
@@ -438,21 +410,21 @@ const CreatePage: NextPage = observer(() => {
...
-    if (isRangeOutcomeEntry(formData.outcomes.value)) {
-      const bounds = [
-        formData.outcomes.value.minimum,
-        formData.outcomes.value.maximum,
-      ];
-      return new Promise(async (resolve, reject) => {
-        await store.sdk.models.createScalarMarket(
-          ...params,
-          bounds,
+    let marketId: number;
```
Robiquet, line 423: Think this can be declared using `const` a couple of lines below where it's assigned

components/liquidity/PoolSettings.tsx
```diff
@@ -59,23 +59,29 @@ const PoolSettings: FC<{
...
-      ...data.slice(rowIndex + 1),
-    ]);
+  const changeOutcomeRow = (amount: string) => {
+    onChange(
+      data.map((row) => ({
+        ...row,
+        amount,
+      }))
+    );
+    // const currentRowData = data[rowIndex];
```
Robiquet, line 69: Is this commented code needed?

components/ui/Avatar.tsx
```diff
@@ -16,15 +16,15 @@ const Avatar = observer(
     zoomed?: boolean;
     size?: number;
   }) => {
+    if (address === "") {
```
Robiquet, line 19: Does this fix this https://github.com/zeitgeistpm/ui/issues/131? If so could you link it?

Robiquet (changes requested):
Looks good, just a few minor things. Looks like the test net node is down at the moment, I'll have a better look when it's back up

Robiquet:
I'm getting an error when creating a categorical market with a pool like this:
image

saboonikhil:
> I'm getting an error when creating a categorical market with a pool like this: image

I encountered same while testing changes on SDK, I believe they have changed the data type of error index from _integer_ to _hex_ value. I have been able to accommodate this by using following tweak: 

https://github.com/zeitgeistpm/tools/blob/bdfe1654d17423fd4dbe2b3361c41a34c528092d/packages/sdk/src/models/index.ts#L310-L313

components/liquidity/PoolSettings.tsx
yornaath: Yeah not really tbh

yornaath:
> > I'm getting an error when creating a categorical market with a pool like this: image
> 
> I encountered same while testing changes on SDK, I believe they have changed the data type of error index from _integer_ to _hex_ value. I have been able to accommodate this by using following tweak:
> 
> https://github.com/zeitgeistpm/tools/blob/bdfe1654d17423fd4dbe2b3361c41a34c528092d/packages/sdk/src/models/index.ts#L310-L313

This should be fixed in the pr now by including your hex extraction fix 👍

pages/create.tsx
yornaath: The assignment is done based on a if else.
```ts
let marketId: number;

    if (!deployPool) {
      marketId = await new Promise(async (resolve, reject) => {
        ...
        return parseInt(await store.sdk.models.createMarket(params));
      });
    } else {
      marketId = await createCategoricalCpmmMarketAndDeployPoolTransaction();
    }
```

Robiquet: Ah sorry, I must have missed the line expand thing

Issue #135, opened 2022-06-24
Change the way subsquid endpoint is persisted


TvrtkoM:
When app loads with default settings it will test if subsquid is available to use and if not persist that state so another time app loads it will not try to use same subsquid setting again.

Change this so the next time app loads, it will still try to use same subsquid endpoint instead of disabling it automatically.

Issue #134, opened 2022-06-24
Account settings for endpoints not showing correct values after refresh


TvrtkoM:
When custom endpoints are set on account settings, when page is refreshed, default values are shown instead of ones from local storage.

TvrtkoM:
related to #135 - if endpoint fails it shouldn't be persisted

PR #143, opened 2022-07-08
Change rpc endpoint selection on settings page

TvrtkoM:
related issuesL #134 #135


lib/util/index.ts
```diff
@@ -2,7 +2,7 @@ import {
   EndpointOption,
   JSONObject,
   SelectOption,
-  SupportedParachain,
+  SupportedParachain
```
yornaath, line 5: I think our prettier should include the rule for trailing commas, makes diffing more smooth

yornaath (changes requested):
LGTM but a couple of trivial things that should be fixed imho.

PR #162, opened 2022-08-01
Support authorised market dispute mechanism 

### LINKED ISSUE #142: Review authorised market dispute mechanism
Currently the app just supports `SimpleDisputes`. For launch, this mechanism will be removed and replaced with the Authorised mechanism. We need to review the app to made sure that it works end to end with the `Authorised` dispute mechanism.

We should also display the authorised wallet address on market page similarly to how we do with the Oracle and Creator.

Robiquet:
[closes #142] Currently awaiting an SDK release, see here: https://github.com/zeitgeistpm/tools/pull/204


Robiquet:
Ready for review but don't merge until runtime is update (needs swap fees enabled)

PR #166, opened 2022-08-02
Format all files

Robiquet:
As we've recently updated the prettier config I've ran it against all files to reduce the diffs in upcoming prs. I've also corrected the name of the prettier config file so that our IDEs pick it up


yornaath (approved):
LGTM

PR #168, opened 2022-08-02
Wallet selection improvements

TvrtkoM:


lib/wallets/index.ts
```diff
@@ -3,7 +3,10 @@ import Decimal from "decimal.js";
 import { KeyringPairOrExtSigner } from "@zeitgeistpm/sdk/dist/types";
 import { KeyringPair } from "@polkadot/keyring/types";
 import { makeAutoObservable, reaction, runInAction } from "mobx";
-import { decodeAddress, encodeAddress } from "@polkadot/util-crypto";
+import {
+  decodeAddress,
+  encodeAddress as polkadotEncodeAddress,
```
yornaath, line 8:
```suggestion
  encodeAddress,
```
No need to alias since there is no name conflict in the file as far as I can tell.

yornaath (changes requested):
LGTM

PR #182, opened 2022-08-15
Fees selection and usage

### LINKED ISSUE #81: Display fees
Malte mentioned that he is adding fees to the runtime. From what I remember fees is hardcoded as 0 in a lot of places of the app, we should update this to the correct value. May need some design changes to make sure it's clear

Robiquet:
[closes #81]


yornaath (changes requested):
From a code standpoint it looks good to me, but I think tests for the core functions should be added like: `calcSpotPrice`, `calcTotalAssetPrice` and `generateSwapExactAmountOutTx`.

Robiquet:
> From a code standpoint it looks good to me, but I think tests for the core functions should be added like: `calcSpotPrice`, `calcTotalAssetPrice` and `generateSwapExactAmountOutTx`.

I think tests for the simple calculations should be straight forward to add as part of this pr. Though I think the transaction generation functions are a bigger piece of work as we need to figure out how to mock the polkadot js api

yornaath:
> generateSwapExactAmountOutTx

Yeah looking at it I think you only need to create a unit test for `calcInGivenOut` and `calcOutGivenIn` since thats the part that does the client side calculation logic before creating the tx.

yornaath (approved):
LGTM

PR #194, opened 2022-08-19
Add banner, optimise images, styling updates

Robiquet:


components/ui/GlitchImage.tsx
```diff
@@ -1,14 +1,15 @@
 import React, { FC, useState } from "react";
 import { motion } from "framer-motion";
 
-const GlitchImage: FC<{ className?: string; alt?: string; src: string }> = ({
+const GlitchImage: FC<{ className?: string; glitchImageSrc: string }> = ({
   className,
-  src,
+  glitchImageSrc,
```
yornaath, line 6: Why the rename of the prop? I think since its basically an image we should use the standard src prop.

yornaath (commented):
A small question, but other than that LGTM

components/ui/GlitchImage.tsx
Robiquet: Yeah fair point, it's kind of duplicating the component name. I will revert it

PR #195, opened 2022-08-19
makets list fixes

TvrtkoM:
requires this PR in sdk https://github.com/zeitgeistpm/tools/pull/215


lib/stores/MarketStore.ts
```diff
@@ -27,6 +27,7 @@ import { JSONObject, MarketOutcome, MarketStatus } from "../types";
 import Store from "./Store";
 import { calcSpotPrice } from "lib/math";
 import { AssetIdFromString } from "@zeitgeistpm/sdk/dist/util";
+import { toHMS } from "components/ui/Countdown";
```
Robiquet, line 30: Don't think this is used

Robiquet (commented):
Looks good! I think it would be cool if the 'My Markets' page reacted to connected wallet changes, if it's easy it would be good to do here if not lets put it on another ticket

Robiquet (approved):

PR #198, opened 2022-08-23
Upgrade SDK to latest 0.7.4 + create market ipfs cluster pinning fail fix

### LINKED ISSUE #150: Settings - in case of faulty rpc url app still tries to connect after reverting
After a number of tries the app will revert to last known configuration that worked or will connect to the testnet. After the revert happened, cnsole still prints  errors for the faulty url e.g. 

```
logger.js?767c:81 2022-07-25 15:51:51          API-WS: DOMException: Failed to construct 'WebSocket': The URL 'ws://127.0.0.1:99444' is invalid.
    at WsProvider.connect (webpack-internal:///./node_modules/@polkadot/rpc-provider/cjs/ws/index.js:509:204)
    at WsProvider.connectWithRetry (webpack-internal:///./node_modules/@polkadot/rpc-provider/cjs/ws/index.js:543:20)
    at eval (webpack-internal:///./node_modules/@polkadot/rpc-provider/cjs/ws/index.js:546:16)

```

### LINKED ISSUE #199: Create market should fail and not create on chain if adding to IPFS cluster fails.
> @sea212 Would you say that the app should make it a hard requirement that the data has been added to the cluster before the market is created on chain?

> @sea212 Definitely.

Blocked:
Needs change in the SDK where createMarket should throw when adding to cluster fails. Right now it just logs.

yornaath:
Upgraded the SDK and accommodated some small changes related to ipfs initialzation.
Seems to work but needs some testing.

Fixes #199 
Fixes #150


yornaath:
@TvrtkoM @Robiquet Cors for ipfs cluster now works and I have verified that it redirects to created market when creating with both pool and without.

Robiquet (approved):

PR #201, opened 2022-08-25
Display cost breakdown

### LINKED ISSUE #174: Make costs clearer on create market page
For launch the bond amounts for market creation will be increasing significantly, the 'transaction cost' for a market will now be at least 500 ZTG. We should make it clearer to the user where the costs are coming from and what will be returned if all goes well.

Cost break down:

- Transaction fee (should be small)
- Advised bond (~200 ZTG) returned if the market is approved or ends before being approved by the committee 
- Permissionless bond (~1000 ZTG) returned if the market isn't deleted by the committee 
- Oracle bond (~200 ZTG) returned if oracle reports the market outcome on time
- Liquidity - can be withdrawn at any time, will collect fees but subject to impermanent loss

Robiquet:
[closes #174]

Bug fixes on create market page
Remove max liquidity as it's no longer required
Derive `minLiqudity` from chain config


yornaath (approved):
LGTM

PR #218, opened 2022-08-30
#215 scalar date markets

### LINKED ISSUE #215: Date market - use scalar markets
Plan is to have a market that has a date as an outcome. Since the categorical markets are not very suitable for this, we could use existing scalar markets with `timestamps` as outcome range and update a code to convert them to real dates for display on the UI

yornaath:
Fixes #215 

- [x] Implement date time input when selecting date type for range outcomes.
- [x] Need metadata changes in the SDK or chain scalar type changes. _Discussing with node/backend team.._
zeitgeistpm/tools#216
- [x] **Blocked** Wait for @saboonikhil to implement `market.scalarType = "number" | "date"` in the subsquid gql type. _Waiting on roll out on `bsr` processor_
- [x] Present dates when applicable in scalar price range slider on market page and market card.


yornaath:
@saboonikhil The SDK integration seems to work as expected and the scalarType shows up on the dev indexer so it seems to be working fine. But will wait to merge this before the change is also rolled out on the `bsr` indexer as the queries will fail when using bsr right now

next.config.js
```diff
@@ -2,4 +2,14 @@ module.exports = {
   experimental: {
     scrollRestoration: true,
   },
+  webpack: (config, { isServer }) => {
```
Robiquet, line 5: What's this change for?

saboonikhil:
> @saboonikhil The SDK integration seems to work as expected and the scalarType shows up on the dev indexer so it seems to be working fine. But will wait to merge this before the change is also rolled out on the `bsr` indexer as the queries will fail when using bsr right now

They're live now!

TvrtkoM:
@saboonikhil I couldn't get `scalarType` to be indexed when using local environment - it's always null regardless of metadata containing `scalarType` field or no.
Market with id 496 on bsr should be with `scalarType: "date"`. Can you please check if this field is indexing correctly?

Issue #204, opened 2022-08-29
Statically render front page


Robiquet:
We should remove the `store.initialized` skeleton from the index page and statically render trending markets and popular categories to reduce load on the infrastructure and ensure users can view the site if the rpc slows down

PR #219, opened 2022-08-30
Statically render front page

### LINKED ISSUE #164: Cache trending markets and market categories
We should cache trending markets and market categories so that they are not loaded again when navigating between the market page and index

Robiquet:
[closes #164 #204]

Currently rerenders every 10mins


Robiquet:
> LGTM man, definitely noticeable 🥳

Thanks yeah seems super quick on my side, just realised I need to make the endpoint string configurable

PR #229, opened 2022-09-07
update next + tailwind

TvrtkoM:


Robiquet (commented):
Looks good, do you think there is any risk merging these changes just before launch? Wondering if it's worth holding off for a week

TvrtkoM:
> Looks good, do you think there is any risk merging these changes just before launch? Wondering if it's worth holding off for a week

Little to no risk to merge, but it's not critical to have and I'm ok with holding it for after launch

Robiquet (changes requested):
Looks like the fonts aren't loading correctly 
image

Robiquet (approved):

Issue #236, opened 2022-09-09
Add proxy support for the authorized account


lsaether:
The UI should check the list of proxied accounts for the authorized account to allow for the proxies to execute on the authorized account's behalf.

PR #245, opened 2022-09-13
allow authority wallet proxy accounts to report

TvrtkoM:
#236


Robiquet (approved):
Cool, I think this is a nice approach for this purpose. Though if we want to use proxies throughout the app we should go about it a more generic way

Issue #306, opened 2022-09-22
Error thrown when user type more than pool amount in tradeslip


Robiquet:
When the user types a number that is higher than the max number of tokens in the pool in the trade slip an error is thrown and the user is unable to trade.

<img width="807" alt="image" src="https://user-images.githubusercontent.com/4950844/191701354-6b64a211-5be7-4460-b78a-03de067d54c1.png">

Issue #251, opened 2022-09-13
Create market page should always be displayed scrolled to the top


Robiquet:
Currently the scroll from other pages is retained and users miss the top field

Robiquet:
Don't think this is fixed. We have this behaviour throughout the app. I think there are some css issues because `window.scroll` never works. I added 
```
    height: 100%;
    width: 100%;
    overflow: auto;
```
to `<main>` which seemed to fix the issue but broke some of the styling

PR #321, opened 2022-09-28
Max ratio trade limits and tradeslip fix

TvrtkoM:
issues #251 #306

### DIFF
--- a/lib/constants.ts
+++ b/lib/constants.ts
@@ -1,3 +1,4 @@
+import Decimal from "decimal.js";
 import resolveTailwindConfig from "tailwindcss/resolveConfig";
 import tailwindConfig from "../tailwind.config";
 import {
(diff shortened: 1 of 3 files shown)


lib/constants.ts
```diff
@@ -11,6 +12,8 @@ import {
 
 export const ZTG = 10 ** 10;
 
+export const MAX_IN_OUT_RATIO = new Decimal(1).div(3).toString();
```
Robiquet, line 15: Did you consider get this from the chain config? Should be on the `swaps` property

lib/stores/TradeSlipStore.ts
```diff
@@ -24,6 +28,7 @@ import {
   generateSwapExactAmountInTx,
   generateSwapExactAmountOutTx,
 } from "lib/util/pool";
+import { IoMdGitMerge } from "react-icons/io";
```
Robiquet, line 31: Don't think this is used

lib/constants.ts
TvrtkoM: Was planning to, but Malte said it's going to be removed from the chain config in near future so I proceeded like this.

Robiquet: Ah good to know

Robiquet (approved):

PR #338, opened 2022-10-08
performance improvements

TvrtkoM:


components/ui/Table.tsx
```diff
@@ -508,7 +508,6 @@ const Table = observer(
 
             {onPaginate ? (
               <Paginator
-                // onIndexChanged={handleIndexChange}
```
Robiquet, line 511: Is this needed?

lib/hooks/useMarketsUrlQuery.ts
```diff
@@ -18,15 +19,20 @@ export const useMarketsUrlQuery = (): MarketListQuery & {
   updateQuery: MarketListQueryUpdater;
 } => {
   const router = useRouter();
-  const rawQuery = router.query;
-
+  const routerPath = router.asPath;
   const query = useMemo(() => {
     try {
-      return parse(rawQuery);
+      const url = parseUri(routerPath);
```
Robiquet, line 25: Have you checked if it's possible to do this with `next/router`?

Robiquet (commented):
Skeletons are no longer appearing below the market cards when infinite scroll is loading. I think we need something so that the user doesn't think they've reached the bottom of the list.
image

lib/hooks/useMarketsUrlQuery.ts
TvrtkoM: Yes, the issue with `router.query` is it is initially empty regardless of query is set initially or no, so this way we don't need to have skeletons on market filters nor the `debouncedQueryChange` state on MarketsList.

components/ui/Table.tsx
TvrtkoM: No, it's a leftover from pagination where you could switch directly to page number, without infinite scroll etc.

TvrtkoM:
> Skeletons are no longer appearing below the market cards when infinite scroll is loading. I think we need something so that the user doesn't think they've reached the bottom of the list. image

At the moment, app only preloads the first page. I'm planning to work on this also as well as displaying market expandable data (besides graph) on preload, but think the app is in a working state now and we should merge the changes so the whole PR doesn't get too big.

TvrtkoM:
Actually we don't have to merge it. but hope it doesn't get too big and complex for the reviewer

Robiquet:
> Actually we don't have to merge it. but hope it doesn't get too big and complex for the reviewer

Is it easy to put back the skeletons or some sort of loader below the market cards when more are loading? I think that's the only thing that's needed

TvrtkoM:
> > Actually we don't have to merge it. but hope it doesn't get too big and complex for the reviewer
> 
> Is it easy to put back the skeletons or some sort of loader below the market cards when more are loading? I think that's the only thing that's needed

Didn't see this tbh, and I don't think it's very simple atm but I'm going to try to fix this today

Robiquet (approved):

PR #371, opened 2022-10-27
changes for new graphql period structure

TvrtkoM:


TvrtkoM:
blocked by https://github.com/zeitgeistpm/tools/issues/232

next.config.js
```diff
@@ -14,14 +14,6 @@ module.exports = withBundleAnalyzer({
       };
     }
 
-    config.externals = [
```
Robiquet, line 17: Is it worth leaving this commented as I think we should have another look at this at some point? 2mb is a big saving

PR #375, opened 2022-10-31
Text editor

Robiquet:


yornaath (approved):
LGMT, seems to work and looks good

PR #384, opened 2022-11-04
Market metadata to contain only cid for market image

TvrtkoM:
This PR contains:
- changing the way market images are saved
- images pinned to ipfs and cid saved instead of big base64 encoded string
- nextjs updated to v13 with all breaking changes fixed

New markets created will have their images pinned to ipfs node, but old markets are still showing the encoded images, until subsquid indexer is changed to reindex old images. The script `scripts/extractMarketImages.ts` extracts all base64 encoded market images from subsquid and saves them to a file.


TvrtkoM:
Needs this PR: https://github.com/zeitgeistpm/tools/pull/234

TvrtkoM:
closed in favor of #401

PR #386, opened 2022-11-07
Fetch prices from market start

### LINKED ISSUE #347: Charts of marktes 0 and 2 not loading properly
The chart of market 0 looks like it only has one data point and doesn't display correctly. The chart of market 2 is stuck loading.

Screenshot 2022-10-13 at 21 13 53
Screenshot 2022-10-13 at 21 14 12

Robiquet:
Need to use market start once subsquid is migrated
[closes #347 ]


Robiquet:
Fix here https://github.com/zeitgeistpm/ui/commit/d33ead9909dd5afeca5b8de450412ff31808a640

PR #390, opened 2022-11-08
Selective SSG 

### LINKED ISSUE #373: Be more specific about what markets are prerendered
We should be more specific about what markets are pre-rendered to reduce build times. Probably, active or proposed or markets that have ended in the last month would be a good selection to pre-render.

Robiquet:
[closes #373 ]


Robiquet:
Simplified the approach a bit to only render markets that have an end date greater that one month ago

Robiquet:
fast builds again!
image

PR #398, opened 2022-11-13
Add notification message as env var

lsaether:
Adds this thing to show up when the env var `NEXT_PUBLIC_NOTIFICATION_MESSAGE` is set in the vercel config.

image

Also my vscode auto modified the CSS file based on the prettier config. We should modify the prettier config if this is not desired behavior.


lsaether:
@TvrtkoM @Robiquet Any idea how to solve this error in pre-rendering markets during the deployment process?

Robiquet:
> @TvrtkoM @Robiquet Any idea how to solve this error in pre-rendering markets during the deployment process?

You might need to pull latest staging into your branch as there have been some recent changes related to the market end

PR #399, opened 2022-11-13
Add featured markets

lsaether:
Adds featured markets to the index page.

Still need to add configuration to make the market ids chosen by env variable instead of being hardcoded.


yornaath (approved):
Code looks solid, added some featured markets to the env and seems to work fine. Approved ✅

PR #400, opened 2022-11-14
nextjs update and remove launch countdown

TvrtkoM:


Robiquet (approved):
Edited the title in case we need to dig out the a launch countdown again

PR #408, opened 2022-11-16
Implement pool details page using sdkv2

### LINKED ISSUE #40: [Liquidity Pools] Pool Volume

### LINKED ISSUE #412: Notifications showing behind modal overlay.
The notification are showing behind the modal overlay, so in some cases like with the buy/sell full setts modals we are using the toast notifications to inform the user that the transaction is underway, included/finalized or errored out. And it is showing behind the gray overlay.

**The notifications should show clearly even when there is a modal.**

yornaath:
**Work done:**

- [x] Show pool details like, volume, value, end date, and status using sdkv2
- [x] Fix volume not showing, fixes #40 
- [x] Show assets, color, ticker etc using sdkv2
- [x] Implement buy and sell full assets using sdkv2.
- [x] _Figure out how to show APR, or replace with other relevant data._ **Went with showing prediction instead.**
- [x] fixes #412 _Notifications showing clearly with modal open_


components/markets/SellFullSetModal.tsx
```diff
@@ -1,139 +1,179 @@
...
-            );
+      notificationStore?.pushNotification("Transacting...", {
+        autoRemove: true,
+      });
+
+      const result = await market
+        .sellCompleteSet({
+          amount: new Decimal(amount).mul(ZTG).toNumber(),
+          signer,
+          hooks: {
```
Robiquet, line 74: We have most of this logic held in `extrinsicCallback` is it possible to reuse that? I think this sort of behaviour should be consistent throughout the app. I also prefer the behaviours provided by `extrinsicCallback` such as a notification displayed after signing and a success message once the transaction is in block

components/markets/SellFullSetModal.tsx
```diff
@@ -1,139 +1,179 @@
...
+              );
+            },
           },
-        }),
-      );
-    };
+        })
+        .asEither();
+
+      if (isRight(result)) {
```
Robiquet, line 95: Some feedback on the sdk. I'm a bit confused by these util functions (`isRight`, `asEither`, `unleft`, etc). They are documented but often refer to other functions in the SDK. For example `unleft` requires and understanding of `right` and `unwrapOr`. Though I don't really understand the concept of left and right in the sdk, is this documented somewhere?

pages/liquidity/[poolid].tsx
```diff
@@ -1,19 +1,26 @@
-import { observer } from "mobx-react";
-import { NextPage } from "next";
-import { useRouter } from "next/router";
-import React, { useEffect, useState } from "react";
-import { ChevronLeft, Info } from "react-feather";
+import { useQuery } from "@tanstack/react-query";
+import {
+  isAvailable,
+  isIndexedData,
+  NA,
```
Comment (Robiquet) [inline pages/liquidity/[poolid].tsx:5]: What's the purpose of `NA`?

pages/liquidity/[poolid].tsx
```diff
@@ -181,46 +186,73 @@ const PoolDetails: NextPage = observer(() => {
...
-                }).format(new Date(marketStore.endTimestamp))
+                }).format(new Date(ends))
               : ""
           }
         />
-        <Pill title="Volume" value="" />
-        <Pill title="Status" value={pool?.pool.status} />
+        <Pill
+          title="Volume"
+          value={`${isAvailable(volume) ? volume : "NA"} ZTG`}
```
Comment (Robiquet) [inline pages/liquidity/[poolid].tsx:198]: Probably 0 makes more sense than "NA"

pages/liquidity/[poolid].tsx
Comment (yornaath) [reply pages/liquidity/[poolid].tsx:5]: NA is for values that arent available in some particular context like pool volume if indexer isnt connected.

yornaath:
@Robiquet Any of these you still want to change or is it mergable?

Robiquet:
> @Robiquet Any of these you still want to change or is it mergable?

Would we good to unify transaction notification behaviour with what we currently have if possible

components/markets/SellFullSetModal.tsx
TvrtkoM: I was wondering the same thing.

TvrtkoM:
The 404 page doesn't work anymore.

components/markets/SellFullSetModal.tsx
yornaath: Made an issue for a refactoring for this. Have to think about the best approach for it

Issue #447, opened 2022-11-23
Create pure <Input /> component conformed to RHF standard.


yornaath:
Create a pure <Input /> component that conforms to standard input props with styling and design according to zeitgeist design and showing error messages.

TvrtkoM:
can this be closed? I think this is done

PR #465, opened 2022-11-28
#447 Refactor tradeslip to use sdkv2, react-query, jotai and rhf

yornaath:
**Work done:**

- [ ] Move state and data fetching from mobx to hooks + jotai atoms
- [ ] Refactor components to use new tradeslip state hooks.
- [ ] Implement data fetching using sdkv2.
- [ ] Implement transaction using sdkv2.
- [ ] Implement form state and validation using RHF.


TvrtkoM:
Is there a need to use TradeSlipStore with new state for tradeslip implemented?

yornaath:
> Is there a need to use TradeSlipStore with new state for tradeslip implemented?

The only thing that uses it is the right drawer component to determine if it should change to tradeslip if items length changes. I can refactor this to use the new jotai tradeslip state.

components/trade-slip/TradeSlipItem.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const prevDebouncedAmountInput = usePrevious(debouncedAmountInput);
+
+  useEffect(() => {
+    if (prevDebouncedAmountInput !== debouncedAmountInput) {
+      put({ ...item, amount: Number(debouncedAmountInput) });
+    }
+  }, [debouncedAmountInput]);
+
+  const onAmountChange = (val: string) => setAmountInput(Number(val));
+  //const onAmountChange = (val: string) => ;
```
Robiquet, line 40: Think this can be deleted

components/trade-slip/TradeSlipItem.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const { put, removeAsset } = useTradeslipItems();
+  const state = useTradeslipItemState(item);
+
+  const [amountInput, setAmountInput] = useState(item.amount);
+  const [debouncedAmountInput] = useDebounce(amountInput, 300);
+  const prevDebouncedAmountInput = usePrevious(debouncedAmountInput);
+
+  useEffect(() => {
+    if (prevDebouncedAmountInput !== debouncedAmountInput) {
+      put({ ...item, amount: Number(debouncedAmountInput) });
```
Robiquet, line 35: I would prefer more verbose but meaningful function names

components/trade-slip/index.tsx
```diff
@@ -1,159 +1,139 @@
...
-              submitTransactions();
-            }}
-            disabled={
-              !tradeSlipForm.isValid ||
-              batchTx == null ||
-              txInProgress ||
-              insufficientZtg
-            }
+            onClick={processTransactions}
+            disabled={isTransacting || tradeslipItems.items.length === 0}
```
Robiquet, line 116: Looks like for validity check has been removed. Is that covered somewhere else?

pages/_app.tsx
```diff
@@ -103,6 +105,9 @@ const MyApp = observer(({ Component, pageProps }) => {
                 <Component {...pageProps} />
               </Layout>
             </DefaultLayout>
+            {process.env.NEXT_PUBLIC_REACT_QUERY_DEVTOOLS === "true" && (
```
Robiquet, line 108: Could you add this to `example.env` so that we can keep track?

components/trade-slip/BuySellButtons.tsx
```diff
@@ -1,124 +1,80 @@
...
-      store.toggleDrawer("right");
+const BuySellButtons = observer(
+  ({ assetId, disabled }: BuySellButtonsProps) => {
+    const store = useStore();
+    const tradeslip = useTradeslipItems();
+    const isDisabled = false;
+
+    if (
+      !IOCategoricalAssetId.is(assetId) ||
+      !IOCategoricalAssetId.is(assetId)
```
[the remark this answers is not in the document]
yornaath: Ah good catch, its supposed to be categorical || scalar. IO means parsing input or output to a valid type. Like with superstruct or zod.

lib/state/tradeslip/tradeslipItemsState.ts
```diff
@@ -0,0 +1,371 @@
...
+        ],
+        queryFn: async () => {
+          if (!enabled || !sdk || !isRpcSdk(sdk)) {
+            return null;
+          }
+
+          const max = (() => {
+            if (item.action === "buy") {
+              const maxTokens = isNA(traderZtgBalance)
+                ? new Decimal(Infinity)
```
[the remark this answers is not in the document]
yornaath: Setting it to 0 means the input cant be changed for users that havent connected their wallet, meaning we dont know their ZTG balance.. But Im changing it to the `tradeablePoolBalance` so that the user can set the input to the maximum allowed to buy from the pool granted they have the ZTG.

components/trade-slip/TradeSlipItem.tsx
yornaath: For the put function?

Robiquet: Yeah

Robiquet (changes requested):
Actually don't merge this, seems like there are still bugs

components/trade-slip/TradeSlipItem.tsx
yornaath: I disagree. The tradeslip items atom is just storage and put is a pretty conventional method name that does either create or update depending on if the item exists. Which is what you want in a context where you dont know if you are creating the item or updating it. https://restfulapi.net/rest-put-vs-post/

Robiquet:
Seems like trade slip items are being removed when I add an amount

Robiquet:
- To receive doesn't change when amount is changed
- To receive amount seems incorrect:

- Amount Input gets cleared when adding another item to the trade slip

Robiquet:
Getting 'Infinity ZBS' on that same market 
image

Robiquet (approved):

PR #470, opened 2022-11-29
Grace fix main

Robiquet:
Same as https://github.com/zeitgeistpm/ui/pull/468 but for main branch


Robiquet:
Think pr is failing to build because Vercel is using preview config. Should be fine once merged

Issue #423, opened 2022-11-21
Replace font Lato with Inter in newly designed components & pages


TvrtkoM:

Robiquet:
https://fonts.google.com/specimen/Inter

PR #479, opened 2022-12-08
#423 replace fonts

robhyrk:
- replaced Lato with Inter
- removed unused fonts
- made Inter default
- used nextJS font optimize to load fonts at build time

### DIFF
--- a/lib/util/fonts.ts
+++ b/lib/util/fonts.ts
@@ -1,10 +1,7 @@
 //font optimization from @next/font
 import { Inter } from "@next/font/google";
 import { Kanit } from "@next/font/google";
-import { Lato } from "@next/font/google";
-import { Roboto } from "@next/font/google";
 import { Roboto_Mono } from "@next/font/google";
-import { Space_Grotesk } from "@next/font/google";
 
 export const inter = Inter({
   subsets: [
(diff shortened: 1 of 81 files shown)


lib/util/fonts.ts
```diff
@@ -0,0 +1,36 @@
+//font optimization from @next/font
+import { Inter } from "@next/font/google";
+import { Kanit } from "@next/font/google";
+import { Roboto_Mono } from "@next/font/google";
+
+export const inter = Inter({
+  subsets: [
```
Robiquet, line 7: This one doesn’t have weights specified, does it load all by default?

lib/util/fonts.ts
```diff
@@ -0,0 +1,36 @@
...
+    "greek-ext",
+    "latin",
+    "latin-ext",
+    "vietnamese",
+  ],
+  variable: "--font-inter",
+});
+
+export const kanit = Kanit({
+  subsets: ["latin", "latin-ext", "thai", "vietnamese"],
```
Robiquet, line 20: I think we just use this for the Zeitgeist logo, can we get away with just the Latin subset?

lib/util/fonts.ts
robhyrk: Yes, normally you do need to specify a range with variable fonts but because it is self hosted by nextJS it loads the range by default.

lib/util/fonts.ts
robhyrk: Good call, I'll remove the other subsets.

TvrtkoM (approved):

Robiquet (approved):
Thanks!

PR #507, opened 2022-12-16
add playwright and some tests

### LINKED ISSUE #494: Playwright testing library
Investigate and setup playwright for e2e testing.

https://playwright.dev/

There seem to be some advantages over using it over cypress e.g. 
- using native async / await syntax instead of `pseudo` synchronous promises inside cypress
- broader browser support
- running multiple browsers using the same test.
- better paralelization

The most notable difference is playwright runs in nodejs instead of browser, which will make setting up tests easier.

TvrtkoM:
closes #494


e2e/errors.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    { path: "/liquidity", testId: "liquidityPage" },
+  ];
+
+  for (const route of testRoutes) {
+    test(`rotue "${route.path}"`, async ({ page, consoleErrors }) => {
+      await page.goto(route.path);
+
+      const element = page.getByTestId(route.testId);
+      await element.waitFor();
+      await page.waitForTimeout(5000);
```
Robiquet, line 17: Looks like `waitForTimeout` should only be used for debugging as it can create flakey tests https://playwright.dev/docs/api/class-page#page-wait-for-timeout

Is there something else we can use instead?

TvrtkoM:
@Robiquet - removed a `waitForTimeout` with setting different lcoators and waiting for `networkidle` load state.

Robiquet (approved):

PR #560, opened 2023-01-10
530 update hero banner

robhyrk:
- created updated hero banner to be used on new home page
- started documenting component on readme (this can be removed if not necessary)
- will require some styling updates once index page is closer to completion


Robiquet (changes requested):
Thanks! Let me know if you want to talk through any of the comments

Robiquet (approved):

PR #544, opened 2023-01-05
New popular categories

Robiquet:
[closes #533] 

Need to add responsive layout once the design is ready

PR #561, opened 2023-01-10
481 update footer

robhyrk:
- created new footer layout with updated breakpoints
- added testing layouts (DemoLayout and index-new) for updated index page components and styles
- added new breakpoints to tailwind config


pages/index-new.tsx
```diff
@@ -0,0 +1,112 @@
+import LearnSection from "components/front-page/LearnSection";
```
Robiquet, line 1: What's this page for?

components/ui/FooterNew.tsx
```diff
@@ -0,0 +1,184 @@
+import React, { FC } from "react";
```
Robiquet, line 1: Is it possible to swap out our current footer with this component?

Robiquet (commented):
Sorry for the delay, let me know if you want to discuss anything

pages/index-new.tsx
robhyrk: I created this temporarily so that the existing home page can remain while I continue adding components to the new home page (index-new) and be able work on the layout. When it's approved and good to go I'll remove the existing index page and rename _index-new_ to _index._

components/ui/FooterNew.tsx
robhyrk: Yes, but it won't respond correctly to breakpoints since the right/left drawers are still there. Might be best to wait for now until they are removed.

pages/index-new.tsx
Robiquet: Only issue I see with this approach is that if anything changes on the index page (I have some changes here: https://github.com/zeitgeistpm/ui/pull/544) we'll need to make sure they are pulled into `index-new` and we won't really have git to help us. Might be easier to build of the components first then compose them once they're ready. Happy for you to make the call though

Robiquet (approved):

pages/index-new.tsx
robhyrk: I see what you're saying. I think I'll keep it though so I can do some testing on the layout itself for now but after I'll update the original index page instead once we are ready to make the switch

PR #567, opened 2023-01-11
Use next-pwa to enable service workers.

yornaath:
Investigating the use of next-pwa and service workers and how it affects performance.


yornaath:
Closing this for now since I see no real benefit to having a service worker or PWA capabilities right now

PR #578, opened 2023-01-11
Revert to using old sdk market extrinsic for buy / sell complete sets

### LINKED ISSUE #573: Use api directly for buying and selling full sets of tokens

TvrtkoM:
close #573


Robiquet (commented):
Code looks good, what's the reason for the change?

TvrtkoM:
> Code looks good, what's the reason for the change?

Currently buying / selling full set doesn't work on main app cause useMarket doesn't return market with methods and button stays disabled all the time

Robiquet (approved):

PR #583, opened 2023-01-13
add index page e2e tests

TvrtkoM:


e2e/index.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    const titleAfterPrev = await heroSlider.locator("h2").textContent();
+    const imageSrcAfterPrev = await image.getAttribute("src");
+
+    expect(titleAfterPrev).not.toBe(titleAfterNext);
+    expect(imageSrcAfterPrev).not.toBe(imageSrcAfterNext);
+  });
+
+  test("popular categories buttons open correct urls", async ({ page }) => {
+    test.setTimeout(100000);
```
Robiquet, line 63: Is this needed?

TvrtkoM: this configures this specific test so it doesn't abort due to timeout, which is 30 seconds by default for each test.
in other words it overrides `timeout` in `playwright.config.ts`, but just for this test

PR #585, opened 2023-01-15
update learn section styles

robhyrk:
- updated "learn section" styles according to latest designs


components/front-page/LearnSection.tsx
```diff
@@ -35,30 +53,30 @@ const LearnCard = ({
...
+          img="/learn/learn-2.png"
         />
         <LearnCard
-          tag="Create"
-          title="Create Markets"
-          description="Learn about the World"
+          tag="Native Currency"
+          title="ZTG Token"
+          description="Tokenomics and Future of ZTG"
           link="https://docs.zeitgeist.pm/docs/learn/market-rules"
```
Robiquet, line 77: Could you find an appropriate link to match the new copy?

Robiquet (commented):
Looks good, just a couple of small things

components/front-page/LearnSection.tsx
```diff
@@ -14,19 +16,35 @@ const LearnCard = ({
...
   return (
     <motion.div
-      whileHover={{ scale: 1.03 }}
+      whileHover={{
+        scale: 1.03,
+        boxShadow: "0px 10px 20px 5px rgba(0,0,0,0.25)",
+      }}
       whileTap={{ scale: 1 }}
-      className={`${className} min-w-[220px] w-full h-[126px] rounded-[10px] p-[15px]`}
+      animate={{}}
```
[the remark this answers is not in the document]
robhyrk: No, it's required because the classes from the individual card props need to be passed through, but I'll change the prop name to avoid confusion though.

components/front-page/LearnSection.tsx
```diff
@@ -1,32 +1,49 @@
...
+  img: string;
+  classes?: string;
 }
 
 const LearnCard = ({
   tag,
   title,
   description,
-  className,
+  classes,
```
TvrtkoM, line 17: think this should have default value set to empty string, otherwise the class would show `undefined` for an element if the parent component doesn't set anything.

components/front-page/LearnSection.tsx
```diff
@@ -1,32 +1,49 @@
...
   return (
     <motion.div
-      whileHover={{ scale: 1.03 }}
+      whileHover={{
+        scale: 1.03,
+        boxShadow: "0px 10px 20px 5px rgba(0,0,0,0.25)",
+      }}
       whileTap={{ scale: 1 }}
-      className={`${className} min-w-[220px] w-full h-[126px] rounded-[10px] p-[15px]`}
+      className={`${classes} w-full rounded-[10px] p-[15px] pr-[20px]`}
```
TvrtkoM, line 28: shouldn't the `${classes}` go at the end of the string so it's possible to override default styles e.g. setting rounded corners to 0, or changing padding. Otherwise we'd need to use `important`

TvrtkoM (changes requested):

PR #589, opened 2023-01-16
532 update market cards

robhyrk:
- updated styles based on new designs
- created reusable HorizontalScroll component
- adjusted logic for cards shown based on container width
- added dynamic widths for cards based on container width

note:
- some styles may not be accurately shown due to new breakpoints and the left/right drawers still in place
- kept old code temporarily in /market-card/_index.tsx and ./markets/_MarketScroll.tsx in case we need to reference/revert back to it (will remove these once index page is approved)
- will also cleanup remaining commented code once index page is approved


components/markets/MarketScroll.tsx
```diff
@@ -1,38 +1,48 @@
...
       "right",
     );
     const { width: containerWidth, ref: containerRef } = useResizeDetector();
-    const cardWidth = 320;
-    const gap = 30;
+    const gap = 16;
+    //calculate cards shown and width based on container width
+    const cardsShown =
+      containerWidth < 716
+        ? 3
```
TvrtkoM, line 27: Think this is a typo. on small screens it shows only one card but scrolls 3 at a time.

components/markets/MarketScroll.tsx
```diff
@@ -1,38 +1,48 @@
...
+        : containerWidth >= 716 && containerWidth < 1183
+        ? 2
+        : 3;
+    // const cardsShown = Math.floor(containerWidth / (gap + cardWidth));
```
TvrtkoM, line 31: can this be removed?

components/markets/MarketScroll.tsx
```diff
@@ -1,38 +1,48 @@
...
     const scrollMin = 0;
     const scrollMax = cardWidth * markets.length + gap * (markets.length - 1);
-    const cardsShown = Math.floor(containerWidth / (gap + cardWidth));
+
     const moveSize = cardsShown * (cardWidth + gap);
 
     useEffect(() => {
       scrollRef.current.scroll({ left: scrollLeft, behavior: "smooth" });
-    }, [scrollRef, scrollLeft]);
+    }, [scrollRef, scrollLeft, cardWidth]);
```
TvrtkoM, line 45: strange behavior in my opinion, when the screen width (and the cardWidth) is changed, the component scrolls the market cards. On smallest breakpoint cards can end up looking like this
image

components/markets/MarketScroll.tsx
```diff
@@ -59,58 +69,36 @@ const MarketScroll = observer(
...
-            </button>
-          </div>
-        </div>
-        <div className="relative">
+      <div ref={containerRef} className="grid sm:grid-cols-2 gap-4 md:gap-6">
+        <h3 className="sm:col-span-1 font-bold text-[28px]">{title}</h3>
+        <HorizontalScroll
+          classes="order-2 sm:order-none"
+          link={link}
+          cta="Go To Markets"
```
TvrtkoM, line 77: Think `cta` should also be passed as a prop to the component. Maybe use a specific type for both `cta` and `link`.

pages/markets/index.tsx
```diff
@@ -35,7 +35,7 @@ const MarketsPage: NextPage<{
           <MarketScroll
             title="Newest Markets"
             markets={newestMarkets}
-            showMarketsLink={false}
+            // showMarketsLink={false}
```
TvrtkoM, line 38: think this comment can be removed now.

TvrtkoM:
there is a console error now when going to `markets/` route
```
Warning: `NaN` is an invalid value for the `minWidth` css style property.
```

TvrtkoM (changes requested):
Looks good. Think it needs some adjustments. Refer to comments above

lsaether:
Can you work with @mk26071980 to define hover behavior on the market row scroll component? Something similar to what was implemented on the new learn cards.

components/markets/MarketScroll.tsx
robhyrk: I made a mistake on the scroll functionality breakpoint; scrolling shouldn't appear until 905px breakpoint. I have made the adjustment now.

The market cards should stack one-by-one until hitting 905px so the logic on cardsShown for < 716 is unnecessary; I have removed it.

components/markets/MarketScroll.tsx
robhyrk: yes

components/markets/MarketScroll.tsx
robhyrk: Yes, actually upon review adding cardWidth to the useEffect is unnecessary. Width prop gets updating just fine without it.

robhyrk:
Thanks for the feedback. I made the changes and added the hover states as per [designs](https://www.figma.com/file/A0OEqjIWy6jL4Rk4kT8ddr/UX%2FUI_Workflow?node-id=1327%3A8329&t=elNlcY7q4Vs6cUHk-0)

TvrtkoM (approved):

PR #599, opened 2023-01-21
512 update home page

robhyrk:


lsaether (approved):
Looks great!

Robiquet (commented):
We should be using `next/image` to display images instead of `background-image:url(/carousel/superbowl.png)` as it will automatically optimise the image for the client. 3.5mb too big to be loading unoptimised. We should also utilise the `blur` property on `next/image` so that there is not a large white space whilst the image is loading

Robiquet:
Is it possible to start with the right drawer closed? Looks like this for me
image

Robiquet (approved):

PR #621, opened 2023-01-24
Subscribe balance events

### LINKED ISSUE #598: Make app reactive to data changes with react query
We need a way to update the data when when important pieces of information change such as prices and balances.

Robiquet:
Subscribe balance events and invalidate cache for relevant accounts
[closes #598 ]


lib/hooks/queries/useAccountAssetBalances.ts
```diff
@@ -38,7 +38,7 @@ export type AccountAssetIdPair = {
   assetId: AssetId;
 };
 
-export const rootKey = "account-asset-balance";
+export const accountAssetBalanceRootKey = Symbol();
```
yornaath, line 41: Symbol will be null, best to use strings or json serializable data

yornaath (approved):
Approved, but convert symbold

PR #622, opened 2023-01-25
Refactor market page asset fetching

Robiquet:
- Begin rendering before before `MarketStore` is initialised
- Use Graphql data to display asset details earlier
- Load prices and 24hr change async 
- Use more efficient method of fetching 24hr change


yornaath (approved):
LGTM

PR #627, opened 2023-01-26
Refactor market prices calculation for reuse

Robiquet:
Refactor market prices calculation into separate function for reuse on portfolio page

### DIFF
--- a/lib/hooks/queries/useMarketSpotPrices.ts
+++ b/lib/hooks/queries/useMarketSpotPrices.ts
@@ -2,13 +2,17 @@ import { useQuery } from "@tanstack/react-query";
 import { isNA, isRpcSdk } from "@zeitgeistpm/sdk-next";
 import Decimal from "decimal.js";
 import { calcSpotPrice } from "lib/math";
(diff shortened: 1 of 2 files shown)


lib/hooks/queries/useMarketSpotPrices.ts
```diff
@@ -25,58 +29,13 @@ export const useMarketSpotPrices = (marketId: number, blockNumber?: number) => {
...
-              basePoolBalance.toString(),
-              baseWeight,
-              balances[index].free.toString(),
-              weight.len,
-              0,
-            );
+      if (isRpcSdk(sdk) && isNA(basePoolBalance) === false) {
+        const spotPrices: MarketPrices =
+          market.status !== "Resolved"
+            ? calcMarketPrices(market, basePoolBalance as Decimal, balances)
```
Robiquet, line 35: Don't think casting is needed here, the `isNA` check should have covered it but it doesn't seem to work

yornaath (approved):
lgtm

PR #637, opened 2023-01-31
604 update cards

robhyrk:
- [x] look into obtaining end date on certain markets on market page
- [x] check what to show when market has no liquidity


components/markets/market-card/index.tsx
```diff
@@ -70,55 +155,112 @@ const MarketCard = ({
...
   volume,
   baseAsset,
   width,
+  tags,
+  endDate,
+  status,
   className = "",
 }: MarketCardProps) => {
-  const [showDetailsOverlay, setShowDetailsOverlay] = useState<boolean>(false);
+  const { data: spotPrices } = useMarketSpotPrices(marketId);
```
Robiquet, line 167: You shouldn't need to do this as we were already displaying the prediction value so the underlying prices must be getting fetched from somewhere. It will vary depending on where the data is coming from, I can help you work through it if you like

components/markets/market-card/index.tsx
```diff
@@ -70,55 +155,112 @@ const MarketCard = ({
...
+                    value="&#x2713; Verified"
+                    classes="bg-green-light text-green"
+                  />
+                )}
+              </div>
+            </div>
+            <MarketCardInfo question={question} />
+            <div className="w-full">
+              {/* don't show if market type is scalar */}
+              {!marketType?.scalar && (
```
Robiquet, line 256: I think we can show something like the `ScalarPriceRange` component where we display the predicted price and the bar will show where that predicted price is relative to the bounds

Robiquet (changes requested):

components/markets/market-card/index.tsx
robhyrk: I'll follow up with Mikhail regarding designs as the current one doesn't quite match the new aesthetics. If I recall correctly I think this was going to be addressed next sprint anyways, but I've setup the existing component for now.

components/markets/market-card/index.tsx
robhyrk: Thanks for the direction here; I was able to sort it out!

lib/util/assets.ts
```diff
@@ -11,7 +11,7 @@ export const getCurrentPrediction = (
     marketType: { categorical?: string; scalar?: string[] };
     categories: { color: string; name: string; ticker: string }[];
   },
-): string => {
+): [string, number] => {
```
Robiquet, line 14: Could this return an object so that we can know what these things are via the property names? `[string, number]` is a bit cryptic

components/markets/MarketsList.tsx
```diff
@@ -109,7 +109,7 @@ const MarketsList = observer(({ className = "" }: MarketsListProps) => {
   const [markets, setMarkets] = useState<
     (IndexedMarket<Context> & {
       outcomes: MarketOutcomes;
-      prediction: string;
+      prediction: string | [string, number];
```
Robiquet, line 112: Are you sure all these union types are needed? I think it could just be changed to the new type that was added

robhyrk: Changing the above (to return an object) sorted this problem out too now as I changed the types here

TvrtkoM:
@robhyrk I'm getting this when running the app
```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `MarketCardTags`.
```

robhyrk:
> @robhyrk I'm getting this when running the app
> 
> ```
> Warning: Each child in a list should have a unique "key" prop.
> 
> Check the render method of `MarketCardTags`.
> ```

thanks, I addressed this in the latest push now

TvrtkoM (approved):
There is one or two console.logs left, but other than that looks goodi IMO

robhyrk:
> There is one or two console.logs left, but other than that looks goodi IMO

awesome thanks! I'll clean those up before merging

Robiquet (approved):

PR #642, opened 2023-01-31
Display reject reason + SDK upgrade

### LINKED ISSUE #504: Display the reject reason for a market
We emit a reason when rejecting a proposed market. It would be nice to display that to the user. Might be difficult because the market gets deleted. Would it help if we didn't delete the rejected market on chain?

Not high priority.

Robiquet:
closes #504


yornaath (approved):
LGTM 👍

PR #654, opened 2023-02-02
Fix ztg prices

### LINKED ISSUE #653: Fix ZTG price fetching
Fix price fetching and ztg price persistance.
add one more endpoint to fetch prices from
https://zeitgeist.api.subscan.io/api/scan/token

store the value in localstorage so it can be used as a fallback if requests fail

TvrtkoM:
close #653


Robiquet (commented):
Looks like we have similar logic in the sdk so not sure if this is needed https://github.com/zeitgeistpm/sdk-next/blob/HEAD/packages/utility/src/ztg/index.ts#L82

TvrtkoM:
@Robiquet I removed the ztgInfo usage from `Store`. Can you review again?

Robiquet (approved):

PR #661, opened 2023-02-03
right drawer fixed overtop of content

robhyrk:


TvrtkoM:
@lsaether can you please see if this is how we want the right nav to be.

PR #688, opened 2023-02-10
fix: Fixes #683. Add top margin 76px so right open drawer does not obstruct struct wallet info

### LINKED ISSUE #683: Account info obstructed by sidebar
when i go to https://staging.zeitgeist.pm/ the top of this horizontal drawer when opened should be lowered so it doesn't obstruct the area where the user can see their account and its balance.

Screen_Shot_2023-02-10_at_5 45 09_pm

i.e. use some css or something like adding `top: 65px` or similar like this so new user's don't get confused about how to switch accounts if they don't realise that sidebar is obstructing their account

Screen Shot 2023-02-10 at 5 50 37 pm

an alternative is to make the arrow larger, as the user may not even realise they opened the sidebar as not all are tech savvy

ltfschoen:
lowers right drawer when opened so it appears like below

Screen Shot 2023-02-11 at 10 40 46 am


Robiquet:
Hey @ltfschoen, thanks for putting this together. We're currently reworking the layout of the app and this change isn't really aligned with where we want to take things so I'm going to close this. Here is a preview of the sort of design we're going for
image

PR #700, opened 2023-02-14
Range input component

TvrtkoM:


components/ui/inputs.tsx
```diff
@@ -413,3 +413,89 @@ export const TextArea: FC<TextAreaProps> = observer(
...
+  step?: string;
+  value?: string;
+  valueSuffix?: string;
+  className?: string;
+  name?: string;
+  onValueChange?: (val: string) => void;
+  onChange?: ChangeEventHandler<HTMLInputElement>;
+};
+
+const RangeInput = React.forwardRef(
```
Robiquet, line 431: Do you think it would be better to put this in a separate file? I'm finding in a bit hard to track down input components because I can't use the file search for them

TvrtkoM: ok

Robiquet (approved):

PR #701, opened 2023-02-15
662 scalar range bar

robhyrk:


components/markets/ScalarPriceRange.tsx
```diff
@@ -6,7 +6,7 @@ import type { ScalarRangeType } from "@zeitgeistpm/sdk/dist/types";
 import moment from "moment";
 
 interface ScalarPriceRangeProps {
-  scalarType: ScalarRangeType;
+  scalarType: string | ScalarRangeType;
```
Robiquet, line 9: What's the reason for the union type here? I thought `ScalarRangeType` should cover it

components/markets/market-card/index.tsx
```diff
@@ -192,62 +216,64 @@ const MarketCard = ({
     baseAsset: baseAsset?.toUpperCase() ?? "ZTG",
   };
 
+  const lower = Number(marketType?.scalar?.[0]) / 10 ** 10;
```
Robiquet, line 219: Could you use the `ZTG` constant instead of `10 ** 10`?

components/markets/ScalarPriceRange.tsx
robhyrk: I thought so too but when I traced it back it seems to be coming from the sdk which has assigned it a type of `string | null`. I can talk to @yornaath and see what might be the issue

robhyrk: I needed to assert `scalarType` to `const ` and that sorted it out

TvrtkoM (approved):

PR #708, opened 2023-02-16
Format compact

### LINKED ISSUE #668: Generalise number formatting
We should have a consistent way for formatting numbers that gives sufficient precision whilst not taking up too much space. I think something like this would work well:

```
new Intl.NumberFormat('default', { maximumSignificantDigits: 3,notation: 'compact'}).format(73987534)
// '74M'

new Intl.NumberFormat('default', { maximumSignificantDigits: 3,notation: 'compact'}).format(0.00903769847561)
// '0.00904'

new Intl.NumberFormat('default', { maximumSignificantDigits: 3,notation: 'compact'}).format(10000)
// '10K'

new Intl.NumberFormat('default', { maximumSignificantDigits: 3,notation: 'compact'}).format(5433)
// '5.43K'
```

Robiquet:
[closes #668 ]


robhyrk (approved):

robhyrk (approved):

Robiquet:
> Looks great! I'll include this in #701 as I'm using the same formatting.
> 
> My only suggestion might be to call the function something like `formatNumCompact` so it's a bit more obvious as to what the function does.

Bit torn on that. With typescript I think it's less important to add information about input/return types in function names as typescript documents that for us. I've made the change though

PR #727, opened 2023-02-21
Trade component

TvrtkoM:


Robiquet:
Thanks for the updates. I noticed the behaviour is a little strange when the user doesn't have a balance of the asset
image

components/trade-slip/TradeSlipItem.tsx
```diff
@@ -1,134 +0,0 @@
-import { isIndexedData } from "@zeitgeistpm/sdk-next";
```
Robiquet, line 1: Could we keep the trade slip components for now? I think there are plans to have something similar so they might be useful

lib/hooks/queries/useTradeItemState.ts
```diff
@@ -0,0 +1,129 @@
...
+    { account: poolAccountId, assetId: item.assetId },
+  ]);
+
+  const poolAssetBalance = new Decimal(
+    poolAssetBalances
+      ?.get(poolAccountId, item.assetId)
+      ?.data?.balance?.free.toString() ?? 0,
+  );
+
+  const query = useQuery(
```
Robiquet, line 70: I think this will need to be reworked as it doesn't look like this will update when prices change. Can be a separate ticket though

components/ui/TransactionButton.tsx
```diff
@@ -54,6 +54,7 @@ const TransactionButton: FC<TransactionButtonProps> = observer(
         onClick={(e) => click(e)}
         disabled={isDisabled()}
         data-test={dataTest}
+        type="submit"
```
Robiquet, line 57: Have you verified that this change doesn't alter the behaviour of any of the other forms in the app?

lib/hooks/trade.tsx
```diff
@@ -0,0 +1,238 @@
...
+    slippage,
+  } = itemState;
+
+  let transaction:
+    | SubmittableExtrinsic<"promise", ISubmittableResult>
+    | undefined;
+
+  if (isRpcSdk(sdk)) {
+    try {
+      if (item.action == "buy") {
```
Robiquet, line 191: I'm not sure if this is quite right. I think the "Exact Amount" should be determined by the last input the user changed. Can be a separate ticket though

lib/hooks/trade.tsx
```diff
@@ -0,0 +1,238 @@
...
+            item.assetId,
+            amountAssetDecimal.toFixed(0),
+            { Ztg: null },
+            minAmountOut.toFixed(0),
+            null,
+          );
+        }
+      }
+    } catch (error) {
+      console.log(error);
```
Robiquet, line 233: I think it's best to throw an error here or not catch at all, otherwise it won't get picked up by react-query. https://tanstack.com/query/v4/docs/react/guides/query-functions#handling-and-throwing-errors

Robiquet (approved):
Thanks!

PR #733, opened 2023-02-22
files for docker and docker compose

TvrtkoM:


docker-compose.yml
```diff
@@ -0,0 +1,9 @@
+version: '3'
+services:
+  web:
+    build: .
+    ports:
+      - "3000:3000"
+    environment:
+      - NODE_ENV=development
+    command: yarn dev
```
Robiquet, line 9: Is it possible to use a production build instead?

Robiquet (approved):

PR #734, opened 2023-02-22
Static side bar

### LINKED ISSUE #606: Static sidebar

robhyrk:
- [x] fix topbar styling to match latest designs
- [x] remove js listeners with css effects
- [x]  topbar opacity?
- [x]  account button skeleton

closes #606


components/account/AccountButton.tsx
```diff
@@ -77,34 +76,50 @@ const AccountButton: FC<{
             )}
           </div>
         ) : (
-          <div className="flex h-ztg-40">
+          <div
+            className={`flex flex-1	items-center justify-end h-full rounded-full cursor-pointer ${
+              pathname === "/" ? "bg-transparent border-white" : "border-black"
```
Robiquet, line 81: Why different colours depending on route? Maybe we could add some opacity to get a similar effect and not have to worry about colours clashing with new banner. Something like twitter:
image

components/account/AccountButton.tsx
```diff
@@ -77,34 +76,50 @@ const AccountButton: FC<{
...
+            }`}
+            onClick={() => {
```
Robiquet, line 83: Connect wallet button needs contrast
image

components/menu/index.tsx
```diff
@@ -50,23 +50,24 @@ const Menu: FC = observer(() => {
           })
           .map((itemKey, idx) => {
             const item = navigationStore.items[itemKey];
-
+            console.log(navigationStore.checkPage(itemKey as any));
```
Robiquet, line 53: Can be removed?

Robiquet (changes requested):

components/menu/MenuItem.tsx
```diff
@@ -54,50 +38,47 @@ export const MenuItem: FC<MenuItemProps> = observer(
...
-          } 
-           rounded-ztg-10 
-          mx-ztg-20
-          `}
+          className={`flex rounded-full p-5 w-[68px] ${className}`}
+          onMouseEnter={() => setIsHovered(true)}
+          onMouseLeave={() => setIsHovered(false)}
           onClick={onClick}
+          style={{
+            border: isHovered
```
Robiquet, line 53: Can we do this will css instead? Whilst the js is loading the menu isn't functional with this approach. Tailwinds `hover:` should work

components/account/AccountButton.tsx
robhyrk: I was going off the designs where the home page has different colours for the topbar components: https://www.figma.com/file/A0OEqjIWy6jL4Rk4kT8ddr/UX%2FUI_Workflow?node-id=2844%3A25907&t=5SVn1YlfoCddYx0T-0

But I do like the idea of adding opacity instead, it would also make it more consistent with the rest of the site. Are you suggesting something more like this?

Screenshot 2023-02-26 at 10 36 20 PM

components/account/AccountButton.tsx
robhyrk: I think it disappears when it's loading...I'll see too adding some sort of skeleton for it instead

components/menu/MenuItem.tsx
robhyrk: I didn't think I could initially, but then I saw that tailwind has something called group effects for child elements so I was able to refactor it

robhyrk: I discovered Tailwind had something called [group effects baed on parent/sibling states](https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state) so I was able to refactor it.

components/account/AccountButton.tsx
Robiquet: No it is there, just back text on black background

components/account/AccountButton.tsx
Robiquet: Yeah kind of though now I'm not sure! Maybe it should kick in further down the page? Though feel free to just leave this for now

components/menu/MenuItem.tsx
Robiquet: oh that's nice, didn't know about that either

components/account/AccountButton.tsx
robhyrk: Found it! I made some changes to the styling and the layout so it isn't so janky when it's loaded

components/account/AccountButton.tsx
robhyrk: I added a transition animation to smooth it out but I'll circle back with Mikhail when I'm working on the mobile menu to get some input from him on this.

Robiquet (approved):
Thanks, looks good!

PR #764, opened 2023-03-01
Add scalar range to market without liquidity

### LINKED ISSUE #314: Can't see scalar range in proposed scalar market
Knowing the scalar range is an important part of the review process, so we need to be able to tell what the range is.

robhyrk:
closes #314

### DIFF
--- a/components/markets/market-card/index.tsx
+++ b/components/markets/market-card/index.tsx
@@ -268,7 +268,7 @@ const MarketCard = ({
                 pool={pool}
                 prediction={prediction}
               />
-            ) : (
+            ) : pool !== null ? (
               <ScalarPriceRange
                 scalarType={scalarType}
                 lowerBound={lower}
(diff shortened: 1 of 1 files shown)


components/markets/market-card/index.tsx
```diff
@@ -268,14 +268,26 @@ const MarketCard = ({
                 pool={pool}
                 prediction={prediction}
               />
-            ) : (
+            ) : pool !== null ? (
```
Robiquet, line 271: Can pool be `undefined`?

robhyrk: I don't think so. When I run the markets query in the graphiql explorer it comes back with either `null` or with values inside the `pool `object. Do you think it's worth adding a check for `undefined` to be safe?

Robiquet (approved):

PR #765, opened 2023-03-01
Markets filters styling

### LINKED ISSUE #502: Styling for market filters

TvrtkoM:
close #502


lsaether:
The images look fuzzy, any idea why? 
image

Robiquet:
> The images look fuzzy, any idea why? image

Yeah I noticed next seems to do this with small images. You can adjust it with the quality prop https://nextjs.org/docs/api-reference/next/image#quality

components/markets/market-filter/index.tsx
```diff
@@ -1,35 +1,126 @@
...
+};
+
+const Option = ({ children, ...props }: OptionProps<MarketFilter>) => {
+  const { data } = props;
+
+  const { activeFilters } = useContext(MarketFiltersContext);
+
+  const isActive = findFilterIndex(activeFilters, data) !== -1;
+
+  const imageUrl = (() => {
```
Robiquet, line 69: I think it would be cleaner for this component to take stuff like `imageUrl` as an input so that it doesn't have to know about the context around it. Is is possible to structure it that way?

TvrtkoM: I don't know how react-select would pass `imageUrl` to component. It is passed to `ReactSelect` on line 205 in this file, but I guess it would be possible if `imageUrl` is part of the option and `data` on line 63 would contain it.
Let me try

Robiquet: Ah ok, I was thinking there might be something awkward like that

Robiquet (approved):

PR #766, opened 2023-03-02
Use default description instead of market description

Robiquet:
Market description is often long and contains html tags


TvrtkoM (approved):

lsaether:
We should change this back to the market description (the beginning of it) once we add sanitation to it.

PR #780, opened 2023-03-03
Use notion as cms

yornaath:
**Work done:**
- [x] Render banners based on notion cms data
- [x] Render featured markets based on cms data
- [ ] Render promoted markets based on cms data. **Blocked:**  _(promoted markets not implemented in the UI yet @Robiquet?)_


robhyrk:
Properties the marketing team may want to control on the banner:

```
    screens: {
      sm: "600px",
      md: "905px",
      lg: "1240px",
      xl: "1440px",
      "2xl": "1728px",
    },
```

- [ ] background position (for each breakpoint)
- [ ]  title text and styles (color, size, font weight, margins, font family)
- [ ]  secondary title text and styles  (color, size, font weight, margins, font family)
- [ ]  CTA primary text, link, and styles (bg color, border color, text color)
- [ ] CTA secondary text, link,  and styles (bg color, border color, text color)
- [ ] image url

yornaath:
@robhyrk I would like to limit the input from the marketing team as much as possible for simplicity.

Regarding button style, it should be a list of presets like blue, black, red, or purple. And we set the text and border color accordingly.

Im not sure we should allow setting font families sizes etc. Or at least then make it a set of presets like large, small or medium.

lsaether:
> @robhyrk I would like to limit the input from the marketing team as much as possible for simplicity.
> 
> Regarding button style, it should be a list of presets like blue, black, red, or purple. And we set the text and border color accordingly.
> 
> Im not sure we should allow setting font families sizes etc. Or at least then make it a set of presets like large, small or medium.

I would limit it even further to just the bare minimum styles and then see what the actual business needs are for adding custom styles on the banner. I personally would only like a single banner template with known constraints that can be met by design / marketing. This is so that we don't need to bother on worrying about making custom styles fit, and can just worry about making a single style template that should look good always.

yornaath:
> > @robhyrk I would like to limit the input from the marketing team as much as possible for simplicity.
> > Regarding button style, it should be a list of presets like blue, black, red, or purple. And we set the text and border color accordingly.
> > Im not sure we should allow setting font families sizes etc. Or at least then make it a set of presets like large, small or medium.
> 
> I would limit it even further to just the bare minimum styles and then see what the actual business needs are for adding custom styles on the banner. I personally would only like a single banner template with known constraints that can be met by design / marketing. This is so that we don't need to bother on worrying about making custom styles fit, and can just worry about making a single style template that should look good always.
Screenshot 2023-03-07 at 09 49 30

Is this acceptable?

Robiquet:
Promoted market design is here but not implemented yet https://www.figma.com/file/A0OEqjIWy6jL4Rk4kT8ddr/UX%2FUI_Workflow?node-id=4306%3A25062&t=0RBkkKyrs1Lt2JJA-0

Robiquet:
May need some adjustments to the plaiceholder stuff after resolving conflicts

Robiquet (approved):

PR #796, opened 2023-03-04
Mobile menu

robhyrk:
- [x] add block number
- [x] fix whitespace issue upon open
- [x] fix logo and hamburger menu colors
- [x] fix menu, remove scroll


robhyrk:
I've removed the top-bar component and placed all menu related components to the menu folder for simplicity. This also helped with keeping the side menu aligned with the logo's relative position. 

I also removed all instances of mob-x stores and used nextJS's built in routing hook for the menu logic

robhyrk:
Also, I tried removing some code related to left/right drawers from store but it was causing build errors since it relied on other components. I was going to remove them but I'm not sure if it's ready to do so; might be worth revisiting and cleaning this up at some point.

components/menu/Navigation.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { FC } from "react";
+import { MenuItem, MenuItemMobile } from "./MenuItem";
+import { useRouter } from "next/router";
+
+const Navigation: FC<{
+  navigation: {};
```
Robiquet, line 6: Is there a better type for this?

layouts/DefaultLayout.tsx
```diff
@@ -52,13 +45,12 @@ const DefaultLayout: FC<PropsWithChildren> = observer(({ children }) => {
             }
           `}
         </style>
-        {process.env.NEXT_PUBLIC_MIGRATION_IN_PROGRESS !== "true" && <Menu />}
```
Robiquet, line 55: Is this correct? How does this app look when `process.env.NEXT_PUBLIC_MIGRATION_IN_PROGRESS !== "true"` evaluates to `true`?

robhyrk: That was a mistake from a merge. Have since removed it.

components/menu/Navigation.tsx
robhyrk: Oops, yeah that was meant as a placeholder.  I have made an update in the latest push.

Robiquet (approved):

PR #841, opened 2023-03-10
change extrinsic type depending on trade state 

### LINKED ISSUE #752: Use last changed input for "Exact amount" on trades
At the moment the trade component use `swapExactAmountOut` for buys and `swapExactAmountIn` for sells

Instead the "exact amount" should be the last value the user entered to an input, so that what they want is correctly expressed in extrinsic form

### LINKED ISSUE #753: Ensure trade component is reactive to prices changes
When the price changes, values in the trade component should update and the extrinsic should be recalculated

TvrtkoM:
close #752
close #753


lib/hooks/queries/useTradeItemState.ts
```diff
@@ -67,13 +69,22 @@ export const useTradeItemState = (item: TradeItem) => {
...
+    poolAssetBalance: poolAssetBalance?.toString(),
+    traderBaseBalance: traderBaseBalance?.toString(),
+    traderAssetBalance: traderAssetBalance?.toString(),
+  };
+
   const query = useQuery(
     [
       id,
       tradeItemStateRootQueryKey,
+      poolAccountId,
```
Robiquet, line 83: Could you add `queryClient.invalidateQueries([id, tradeItemStateRootQueryKey, account]);` to `useSubscribeBlockEvents`? That way this query will get invalidated when the pool account balance changes

lib/hooks/queries/useTradeItemState.ts
```diff
@@ -122,6 +136,8 @@ export const useTradeItemState = (item: TradeItem) => {
         !!poolAssetBalance &&
         !!wallets.activeAccount?.address,
       keepPreviousData: true,
+      refetchOnWindowFocus: false,
+      refetchOnMount: false,
```
Robiquet, line 140: Why should refetching be prevented in this case?

lib/hooks/queries/useTradeItemState.ts
TvrtkoM: it get's invalidated when balances change without this, since `balances` are part of the query key (line 85). Is this really necessary. I tested this with app opened on two tabs, made a trade on one tab and trade values change on the second tab

yornaath (approved):

PR #854, opened 2023-03-15
Heading component

### LINKED ISSUE #757: Move the how-to section on index to the bottom of the page

### LINKED ISSUE #805: Inconsistent headers on pages
Different font sizes, spacing, and html tags used for header/titles 

closes #812

### LINKED ISSUE #812: User address cut off on mobile - Avatars and Badges page
Screenshot 2023-03-06 at 12 51 58 PM

robhyrk:
- [x] replace existing headers with new headers component
- [x] confirm with design regarding size types for headers
- [x] fix mobile styles activating on smaller breakpoints

closes #812 
closes #805 
closes #757


robhyrk:
Note: due to how tailwind compiles I was unable to create DRYer method of getting sizes for each breakpoint

components/ui/Heading.tsx
```diff
@@ -0,0 +1,329 @@
+import { PropsWithChildren } from "react";
+
+type fontSize =
```
TvrtkoM, line 3: just a remark that capitalized names for types are used through the app. this is internal type so it doesn't matter in this case.

components/ui/Heading.tsx
```diff
@@ -0,0 +1,329 @@
...
+  size,
+  xxl,
+  xl,
+  lg,
+  md,
+  sm,
+  className = "",
+  children,
+}: PropsWithChildren<HeadingProps>) => {
+  const getSmallSize = (size: fontSize) => {
```
TvrtkoM, line 41: `getSmallSize`, `getMediumSize` ... functions could be moved out of function body since they don't use any variable or prop from the component. They are also almost the same with difference in class name prefixes `sm:`, `md:` etc. so it can be one function with arguments `(size: fontSize, prefix?: string)`

TvrtkoM (changes requested):
I don't mind having a component for heading, but we could also just add default styling for h1, h2, h3, h4, h5, h6 tags somewhere in `styles` and use html tags. e.g.

```
h1 {
    @apply text-2.5xl md:text-2.5-xl xl:text-2.5xl ...
}
h2 {
    ...
}
```

then we could use `className` where needed to override font size.

but regarding this PR, I'd move the `getSize` functions out of function body. This can also be just one function that has additional optional class `prefix` argument.

TvrtkoM:
I accidentally approved this instead other PR. oops

PR #855, opened 2023-03-15
Carousel improvements

### LINKED ISSUE #747: Carousel should end on last image, not restart the cycle
The carousel should end on the final image instead of cycling through.

### LINKED ISSUE #748: Carousel should change image at a slower pace
... or not change at all without user input. I would put it on 10 or 15 second cycles.

yornaath:
**Improved carousel behavior:**
- Subtler banner change animation
- 15 second pr banner "autoplay"
- Staggers the next animation by 45 sec on each user interaction _(prev, next select, crumb)_

Fixes #748 
Fixes #747


yornaath:
> Seems like there is an additional flash when loading the page, is that intentional?
> 
> I've noticed that before the images are cached, when scrolling to the next image the flash happens but original image sticks around. I wonder if the `preload` property on `next/image` could help. This issue existed before these change though

Yeah that issue has been around for a while, should I fix it on this pr? "Carousel improvements" is pretty general so fits the bill.

Robiquet:
> > Seems like there is an additional flash when loading the page, is that intentional?
> > I've noticed that before the images are cached, when scrolling to the next image the flash happens but original image sticks around. I wonder if the `preload` property on `next/image` could help. This issue existed before these change though
> 
> Yeah that issue has been around for a while, should I fix it on this pr? "Carousel improvements" is pretty general so fits the bill.

Yeah that would be good

lib/hooks/slides.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export const useSliderControls = (props: UseSlidesProps): UseSliderControls => {
+  const [currentSlide, setCurrentSlide] = useState<number>(
+    props.currentSlide ?? 0,
+  );
+
+  const pauseTimerRef = useRef<NodeJS.Timeout | null>(null);
+  const [paused, setPaused] = useState(false);
+
+  const prev = (userOrigin?: boolean) => {
```
robhyrk, line 27: I think the names of the function are reversed: `prev` seems to move to slider forward and vice-versa with `next`

The autoplay function didn't seem to work correctly but when I switched the names it seemed to work just fine.

components/hero-slider/HeroSlider.tsx
```diff
@@ -15,22 +15,22 @@ const HeroSlider = ({
...
-    if (banners.length > 1) {
-      const ref = setTimeout(() => {
-        setAnimate(true);
-        moveSlider("next", currentSlide, setCurrentSlide, slidesLength);
-      }, 10000);
-      return () => {
-        clearTimeout(ref);
-      };
+    if (isNumber(prevSlide) && prevSlide !== slider.currentSlide) {
+      console.log("set animate");
```
robhyrk, line 30: Can be removed?

robhyrk (changes requested):

lib/hooks/slides.tsx
yornaath: Ah yeah! Good catch haha.

Issue #90, opened 2022-05-05
Enable transactions on mobile devices


Robiquet:
We should make basic transactions possible with Nova wallet.

Requires design work

PR #861, opened 2023-03-15
Nova wallet integration

Robiquet:
Closes [#90 ]


components/account/AccountButton.tsx
```diff
@@ -31,9 +31,16 @@ const AccountButton: FC<{
   const [hovering, setHovering] = useState<boolean>(false);
   const [showOnboarding, setShowOnboarding] = useState(false);
   const [showGetZtgModal, setShowGetZtgModal] = useState(false);
+  const isNovaWallet: boolean =
+    //@ts-ignore
```
yornaath, line 35: Any way to fix types so we dont have to ignore?

yornaath (commented):
Would like to avoid ts ignores, but other than that lgtm

TvrtkoM (approved):

PR #867, opened 2023-03-16
Remove disable main scroll when cursor is over filters

### LINKED ISSUE #793: User cannot scroll when cursor is above market filters
Screenshot 2023-03-03 at 1 36 10 PM

TvrtkoM:
close #793


Robiquet (approved):
Easy fix in the end! :)

PR #885, opened 2023-03-20
Endpoint select options from .env var

### LINKED ISSUE #771: Endpoint switch should only allow RPC connections to the same chain id that's set for SSR

TvrtkoM:
close #771


lib/hooks/useSdkv2.ts
```diff
@@ -36,27 +26,30 @@ export type UseSdkv2 = [
...
       if (sub && prevId && id !== prevId) {
         setTimeout(() => {
           init.cache.delete(prevId);
           sub.unsubscribe();
         }, 500);
       }
 
-      const sdk$ = init(store);
+      const sdk$ = init(endpointVals, graphQlEndpoint);
       //@ts-ignore todo: adjust type in sdk
```
yornaath, line 49: Can you link the type error you get here? Want to fix before merging.

pages/markets/[marketid].tsx
```diff
@@ -38,23 +34,22 @@ import { AlertTriangle } from "react-feather";
...
+import { filters } from "components/ui/TimeFilters";
 
 const QuillViewer = dynamic(() => import("../../components/ui/QuillViewer"), {
   ssr: false,
 });
 
 export async function getStaticPaths() {
-  const url = process.env.NEXT_PUBLIC_SSR_INDEXER_URL;
-  const client = new GraphQLClient(url);
+  console.log(graphQlEndpoint);
```
Comment (yornaath) [inline pages/markets/[marketid].tsx:51]: Maybe remove logging here

yornaath (changes requested):

lib/hooks/useSdkv2.ts
TvrtkoM: image

yornaath (approved):

PR #917, opened 2023-03-23
#891 'Liquidity only' filter not working

### LINKED ISSUE #891: 'Liquidity only' filter not working when resolved market filter enabled
-doesn't happen when only the 'Active' filter is on. Does happen with 'Active' and 'Resolved' are both on.
-consider removing 'dust' balances keeping markets on this filter

image
image

robhyrk:
closes #891

### DIFF
--- a/lib/hooks/queries/useInfiniteMarkets.ts
+++ b/lib/hooks/queries/useInfiniteMarkets.ts
@@ -86,13 +86,10 @@ export const useInfiniteMarkets = (
         status_in: statuses.length === 0 ? undefined : statuses,
         tags_containsAny: tags.length === 0 ? undefined : tags,
         pool_isNull: withLiquidityOnly ? false : undefined,
-        pool: withLiquidityOnly
-          ? { ztgQty_gt: 0 }
-          : undefined && currencies.length === 0
-          ? undefined
-          : {
-              baseAsset_in: currencies,
-            },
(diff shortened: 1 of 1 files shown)


robhyrk:
I'm not sure how much ZTG would be considered a "dust" balance so the `ztgQty_gt` can be adjusted accordingly. For now I left it at `0`

lib/hooks/queries/useInfiniteMarkets.ts
```diff
@@ -86,12 +86,13 @@ export const useInfiniteMarkets = (
         status_in: statuses.length === 0 ? undefined : statuses,
         tags_containsAny: tags.length === 0 ? undefined : tags,
         pool_isNull: withLiquidityOnly ? false : undefined,
-        pool:
-          currencies.length === 0
-            ? undefined
-            : {
-                baseAsset_in: currencies,
-              },
+        pool: withLiquidityOnly
```
yornaath, line 89: I dont belive this is correct since you are either just filtering for pools that have liquidity or for currencies. Although filtering by ztgQuantity wouldnt make sense if the base asset is like aUSD, but I guess ztgQuantity will have to be renamed to baseAssetQuantity.

So I actually think you can leave it like this for now since we only have one base asset.

yornaath (approved):

PR #937, opened 2023-03-27
#927 refactor notification state

### LINKED ISSUE #927: Refactor `NotificationStore` to using jotai atoms.
Encapsulate state, methods and logic for the NotificationStore in jotai atoms + hooks instead of a mobx store class.

_Another approach is to use only declarative components + portal to render notifications in the same way we do modals. But if we need conditional logic for the notifications(as quoted below) it might be easier to have some sort of global state atoms for notifications._

> Yeah but I mean when the user initiates the transaction, they can click outside the modal while it is transacting thereby closing it so they wont see the success message inside the modal on completion. But the notification still appears. I think ideally the UI would recognize if the modal was closed during a transaction and only show notifications on success/error if it was closed.

### LINKED ISSUE #938: Remove unused ExchangeStore and NavigationStore
These stores are unused and can be removed.

yornaath:
**Work done:**
- [x] Refactor `NotificationsStore` from mobx to using state atoms and hooks.
- [x] Implement state and logic using `valtio` over `jotai`. _Reasoning will follow below._
- [x] Refactor existing `jotai` usage to `valtio`.
- [x] Remove unused `NavigationStore` and `ExchangeStore` from `Store`

**Why I chose valtio over jotai:**
- They have very similar apis so refactoring was quite simple.
- Valtio supports mutating state outside of react so implementing `coroutines` that mutate state is much easier. Like the new implementation of the timer logic in the new `state/notifications` module.
- Since we can subscribe to and set state outside of react with valtio its easier to incrementally refactor existing mobx stores when they are mutually reactive/coupled.

Fixes #927 
Fixes #938


TvrtkoM:
There is this warning appearing in terminal after I run `yarn dev`
```
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
```
It doesn't appear on `staging`. Do you know where does that come from?

TvrtkoM (changes requested):
left a comment about useLayoutEffect usage

PR #946, opened 2023-03-30
#298 Display scalar or categorical on market page

### LINKED ISSUE #298: Proposed markets don't show if scalar or category which is hard for advisory comittee.
In order for the advisory committee to approve markets they need to know whether they are scalar or category.

robhyrk:
closes #298

### DIFF
--- a/components/markets/ScalarPriceRange.tsx
+++ b/components/markets/ScalarPriceRange.tsx
@@ -24,7 +24,6 @@ const ScalarPriceRange = observer(
     status,
   }: ScalarPriceRangeProps) => {
     const { width, ref } = useResizeDetector();
-    status = "Proposed";
     const shortPercentage = 1 - shortPrice;
     const longPercentage = longPrice;
     const averagePercentage = (shortPercentage + longPercentage) / 2;
@@ -84,7 +83,7 @@ const ScalarPriceRange = observer(
 
     return (
       <div ref={ref}>
-        <div className="relative top-1.5 ">
+        <div className="relative top-1.5">
           <div className="flex justify-between">
             <div className="flex flex-col justify-start">
               <span className="mb-2.5 text-sm text-blue">{lower}</span>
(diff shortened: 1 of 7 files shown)


pages/markets/[marketid].tsx
```diff
@@ -157,7 +157,7 @@ const Market: NextPage<{
     ? new Decimal(indexedMarket?.pool?.volume).div(ZTG).toNumber()
     : 0;
   const subsidy = marketSdkv2?.pool?.poolId == null ? 0 : pool?.liquidity;
-
+  console.log(indexedMarket);
```
Comment (TvrtkoM) [inline pages/markets/[marketid].tsx:160]: please remove console.log here

TvrtkoM (changes requested):
Can we adopt ScalarPriceRange component to show the scalar range on market page?
There is a comment on the accompanied issue:
```
There is no way to tell what the scalar range for a market is which is quite important for advisory committee. For example: If we were to approve the market seen above, we would have no way of checking if the top of the range is actually higher than the current number of Youtube subscribers.
```

robhyrk:
> Can we adopt ScalarPriceRange component to show the scalar range on market page? There is a comment on the accompanied issue:
> 
> ```
> There is no way to tell what the scalar range for a market is which is quite important for advisory committee. For example: If we were to approve the market seen above, we would have no way of checking if the top of the range is actually higher than the current number of Youtube subscribers.
> ```

I think part of this has already been addressed previously since the scalar range bar is shown on scalar markets below the table section.

TvrtkoM:
> I think part of this has already been addressed previously since the scalar range bar is shown on scalar markets below the table section. [Example](https://staging.zeitgeist.pm/markets/139). Just takes a little bit of time to load so I will add a skeleton to it.

It is not showing for proposed markets and issue #298 is about importance of showing scalar range for proposed markets. it can be queried like this 
```
query MyQuery {
  markets(where: {marketId_eq: 593}) {
    marketType {
      scalar
    }
  }
}
```

components/markets/ScalarPriceRange.tsx
```diff
@@ -91,42 +92,49 @@ const ScalarPriceRange = observer(
               <span className="mb-2.5 text-sm text-red">{upper}</span>
             </div>
           </div>
-          {showShortAndLongPrices && (
+          {/* TODO: check if this can be removed */}
```
robhyrk, line 95: Can this be removed (also line 130?)? Seems like it's no longer used but not exactly sure what it does

TvrtkoM: it used to show horizontal lines around the actual current prediction in the middle when `shortPrice + longPrice` is < 0.97, but now it seems this sum is always very near 1 so it doesn't show. Just leave it commented out for now.

TvrtkoM (approved):

PR #974, opened 2023-04-05
Fix card widths on initial load

### LINKED ISSUE #957: Fix market card width on initial page load
market card goes to 100% width on horizontal scroll components during initial load

robhyrk:
refactored to use CSS breakpoints for calculating card widths

closes #957


components/markets/MarketAssetDetails.tsx
```diff
@@ -129,6 +129,7 @@ const MarketAssetDetails = observer(
                     )
                   : null,
               change: priceChange,
+              poolId: market.pool.poolId,
```
Robiquet, line 132: Is this needed?

components/ui/Table.tsx
```diff
@@ -14,6 +14,7 @@ import Avatar from "./Avatar";
 import { range } from "lodash";
 import { useIsOnScreen } from "lib/hooks/useIsOnScreen";
 import { useZtgInfo } from "lib/hooks/queries/useZtgInfo";
+import { getCurrentPrediction } from "lib/util/assets";
```
Robiquet, line 17: Think this can be removed too

Robiquet (commented):
Looks good, just need to remove the unused code

components/markets/MarketAssetDetails.tsx
robhyrk: Accidentally popped some stashed code here. Will remove.

TvrtkoM (approved):

PR #986, opened 2023-04-05
Playwright learn section snapshot

TvrtkoM:


e2e/index.spec.ts
```diff
@@ -29,6 +29,15 @@ test.describe("index page", () => {
     }
   });
 
+  test("learn section corresponds the snapshot", async ({ page }) => {
+    const indexPage = new IndexPage(page);
+    await indexPage.goto();
+
+    expect(await indexPage.learnSection.screenshot()).toMatchSnapshot(
```
Robiquet, line 36: How do references work here? Where does `learnSection` come from?

Robiquet (approved):
Looks cool

e2e/index.spec.ts
TvrtkoM: `indexPage.learnSection` is Locator from IndexPage class in `lib/index.page.ts`. First time this runs test fails but creates a snapshot, and on next runs it will compare screenshot with saved snapshot. there is playwright option `--updateSnapshot` which will replace image when design changes

PR #989, opened 2023-04-05
Price in $ shows NaN until ZTG price is fetched

### LINKED ISSUE #944: Price in $ shows NaN until ZTG price is fetched
It should show skeleton

robhyrk:
closes #944


components/ui/Table.tsx
```diff
@@ -108,6 +108,10 @@ const isAmountInput = (cellValue: CellValue): cellValue is Amount => {
   return (cellValue as Amount).onChange !== undefined;
 };
 
+const hasZTGPriceLoaded = (ztgPrice: number) => {
```
TvrtkoM, line 111: result from `useZtgInfo` could be used to check if prices are loaded instead of this.
this hook returns `isLoading` and `isLoadingError` that we could use so if `isLading` is true show Skeleton and if `isLoadingError` is true we could show something different but skeleton is also ok.

refer to `useQuery` documentation - https://tanstack.com/query/v4/docs/react/reference/useQuery

TvrtkoM (changes requested):
please see comment

TvrtkoM (approved):

PR #1003, opened 2023-04-10
657: Display total cost of position on portfolio page

### LINKED ISSUE #657: Display average entry price / total cost of position on the portfolio page
Using the transaction history we should be able to calculate how much a user spent to acquire a token. That way they can know if they are in profit or not

Not sure what is best "Average entry price" or "Total Cost" or something else?

robhyrk:
closes #657 and #1002


lib/hooks/queries/useTransactionHistory.ts
```diff
@@ -68,7 +70,7 @@ export const useTransactionHistory = (address: string) => {
       if (isIndexedSdk(sdk) && address) {
         const { historicalAssets } = await sdk.indexer.client.request<{
           historicalAssets: {
-            assetId: string;
+            assetId: string | AssetId;
```
Robiquet, line 73: Are you sure this can be both types?

lib/hooks/queries/usePortfolioPositions.ts
```diff
@@ -427,13 +434,38 @@ export const usePortfolioPositions = (
         color = "#DF0076";
       }
 
+      const totalCost = transactionHistory
+        .filter((transaction) => {
+          const scalarTransaction = transaction?.assetId["ScalarOutcome"];
```
Robiquet, line 439: I think there are some utils in the new sdk for this sort of stuff. @yornaath could you point us in the right direction?

Robiquet (commented):
I am wondering what we should do if a user sells an asset? Should the total cost stay the same or be reduced by the amount they sold. Not sure if we actually have that information at the moment

lib/hooks/queries/useTransactionHistory.ts
robhyrk: I was testing something out and forgot to remove this. I think it can stay as `string`

robhyrk:
Blocked due to missing transaction info. [Issue opened on subsquid.](https://github.com/zeitgeistpm/zeitgeist-subsquid/issues/340)

Robiquet:
I'm seeing negative total cost on some positions, why's that? 
image

Robiquet (commented):
This feature doesn't seem to work for the Subsidy section so we should probably remove it from there

robhyrk:
> I'm seeing negative total cost on some positions, why's that? image

Did you check if useTradeHistory displayed all the trades for those markets/assets? One of the issues I mentioned was surrounding some missing transactions and that resulted in negative values being calculated since it only showed sells.

The other issue I'm realizing is that `total cost` may not be a good measure because it can result in negative values anyways - which is actually good because it reflects a gain but may be confusing to the user. For example: if I purchase an asset and the price of that asset goes up and then I sell most of that asset then the total cost of acquiring that asset would become negative. 

Example:
10 units of asset `yes `is purchased for 2 ZTG. Cost = 20 ZTG
Asset `yes` price increases to 3 ZTG
8 units of asset `yes` is sold for 3 ZTG. Cost = -24 ZTG
Total cost = -4 ZTG

Average price per unit then that may be a better reflection of the cost of acquiring that asset. Alternatively we could show both `total cost` and `average cost` and include a tooltip that explains how each is calculated.

Robiquet:
> > I'm seeing negative total cost on some positions, why's that? image
> 
> Did you check if useTradeHistory displayed all the trades for those markets/assets? One of the issues I mentioned was surrounding some missing transactions and that resulted in negative values being calculated since it only showed sells.
> 
> The other issue I'm realizing is that `total cost` may not be a good measure because it can result in negative values anyways - which is actually good because it reflects a gain but may be confusing to the user. For example: if I purchase an asset and the price of that asset goes up and then I sell most of that asset then the total cost of acquiring that asset would become negative.
> 
> Example: 10 units of asset `yes `is purchased for 2 ZTG. Cost = 20 ZTG Asset `yes` price increases to 3 ZTG 8 units of asset `yes` is sold for 3 ZTG. Cost = -24 ZTG Total cost = -4 ZTG
> 
> Average price per unit then that may be a better reflection of the cost of acquiring that asset. Alternatively we could show both `total cost` and `average cost` and include a tooltip that explains how each is calculated.

Thanks that makes sense, what do you think about showing profit/loss (current value - how much they spent)? I think that would ultimately be what the user wants to know

robhyrk:
> > > I'm seeing negative total cost on some positions, why's that? image
> > 
> > 
> > Did you check if useTradeHistory displayed all the trades for those markets/assets? One of the issues I mentioned was surrounding some missing transactions and that resulted in negative values being calculated since it only showed sells.
> > The other issue I'm realizing is that `total cost` may not be a good measure because it can result in negative values anyways - which is actually good because it reflects a gain but may be confusing to the user. For example: if I purchase an asset and the price of that asset goes up and then I sell most of that asset then the total cost of acquiring that asset would become negative.
> > Example: 10 units of asset `yes `is purchased for 2 ZTG. Cost = 20 ZTG Asset `yes` price increases to 3 ZTG 8 units of asset `yes` is sold for 3 ZTG. Cost = -24 ZTG Total cost = -4 ZTG
> > Average price per unit then that may be a better reflection of the cost of acquiring that asset. Alternatively we could show both `total cost` and `average cost` and include a tooltip that explains how each is calculated.
> 
> Thanks that makes sense, what do you think about showing profit/loss (current value - how much they spent)? I think that would ultimately be what the user wants to know

Yeah I think so too. I find myself often checking average price too when I buy/sell assets like crypto so I think including PnL and average price is the way to go. I'll update this PR to include those features.

Robiquet:
Just thinking... you can also acquire/lose tokens via liquidity provision and minting/burning full sets. So I wonder if some of the issues are related to that

robhyrk:
I think it would helpful for the user to get an explanation of the terms via a tooltip; I'll open a separate issue for that.

Lastly, I just need to refactor some code so that the table columns conditionally loaded whether or not it's in the subsidy section.

Robiquet (approved):
Thanks, looks good. I don't think we should abbreviate the column headers though as I don't think people are going to understand 'rPnL' and 'uPnl'

Issue #940, opened 2023-03-29
Refactor `Wallets` store to valtio 


yornaath:
Refactor the Wallets state and logic using mobx into using valtio.

PR #1005, opened 2023-04-12
#940 refactor wallets store to use jotai

yornaath:
**Work done:**
- Replace mobx implementation of wallet state and logic with ~~valtio~~ **jotai**
- Hook interface is allmost identical to the store interface.
- Created a `persistentAtom` utility that has the correct persisted state on load and supports migrations.
- Replaced `activeBalance` on the wallet state with usage of the `useZtgBalance` hook in components that needs it.

> @Tvrtko Not sure if you catched all our discussions about state 😅  But I think we have landed back on jotai over valtio. The main reason being that valtio proxies(like stores) are complicated to work with and wrap its properties in proxies that break some values like Decimals and the sdk if stored in proxies. It handles primitive values that are json serializable well, but if we store instances of classes etc they break some times.
The main reason for choosing valtio was that I didnt think you could update jotai state outside of the react tree, which limits some patterns of usage that can be handy. But in jotai v2 we can update state outside.


lib/state/wallet.ts
```diff
@@ -0,0 +1,340 @@
...
+/**
+ * List of supported wallets.
+ */
+export const supportedWallets = [
+  new PolkadotjsWallet(),
+  new SubWallet(),
+  new TalismanWallet(),
+];
+
+let accountsSubscriptionUnsub: VoidFunction | null = null;
```
Robiquet, line 144: Does this need to be here? Or just inside `enableWallet `

lib/state/wallet.ts
```diff
@@ -0,0 +1,340 @@
...
+ * Subscribe to user location and vpn changes and
+ * disconnect the wallet if location isnt allowed or vpn is used.
+ */
+store.sub(userLocationDataAtom, async () => {
+  const data = await store.get(userLocationDataAtom);
+  if (
+    (store.get(walletAtom).connected && !data?.locationAllowed) ||
+    data?.isUsingVPN
+  ) {
+    store.set(walletAtom, (state) => {
```
Robiquet, line 255: Might be nice to use `disconnectWallet` if it's doing the same thing

Robiquet (commented):
Code looks good. What have you tested to ensure nothing has regressed?

components/assets/AssetActionButtons/DisputeButton.tsx
```diff
@@ -21,7 +21,7 @@ const DisputeButton = observer(
     market: Market<IndexerContext>;
     assetId: MarketOutcomeAssetId;
   }) => {
-    const [sdk] = useSdkv2();
+    const [sdk, _] = useSdkv2();
```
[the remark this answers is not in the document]
yornaath: Oh yeah I can remove that, you can use underscore when you dont need one of the params and it wont complain about unused variable, but you can also just omit it.

Both of these works and wont complain.
```ts
const [_, setState] = useState(1)
const [, setState] = useState(1)
``

But:
```ts
const [state, setState] = useState(1)
```
will complain if `state` isnt used anywhere.

lib/state/wallet.ts
yornaath: No this needs to be here since `enableWallet` needs to clear any existing subscription when another wallet is enabled.

lib/state/wallet.ts
yornaath: `disconnectWallet` is only usable by the hook, and if I create one outside the hook that uses the store and then use it in the hook there are issues with state synchronization. I can explain on call, but will leave this for now.

TvrtkoM:
Tried using VPN with location set to US and wallet is still enabled. There is also en error in terminal after running `yarn dev`
```
error - unhandledRejection: TypeError: Only absolute URLs are supported
    at getNodeRequestOptions (/home/tvrtko/programming/zeitgeist/ui/node_modules/next/dist/compiled/node-fetch/index.js:1:63531)
    at /home/tvrtko/programming/zeitgeist/ui/node_modules/next/dist/compiled/node-fetch/index.js:1:65062
    at new Promise (<anonymous>)
    at fetch (/home/tvrtko/programming/zeitgeist/ui/node_modules/next/dist/compiled/node-fetch/index.js:1:64996)
    at global.fetch (/home/tvrtko/programming/zeitgeist/ui/node_modules/next/dist/server/node-polyfill-fetch.js:39:16)
    at Object.queryFn (webpack-internal:///./lib/hooks/useUserLocation.ts:29:36)
```

Robiquet:
> Still looks like there is an issue with not blocking unauthorised locations. Though it looks like the bug is already in staging. When I'm using a US VPN I'm getting blocked on https://app.zeitgeist.pm but not https://test.staging.zeitgeist.pm

Actually this could just be a config thing. Preview config doesn't block any countries

TvrtkoM:
a bug:
1. remove entry `wallet-user-config` from local storage
2. refresh
3. click `conenct wallet` button
4. select extension
5. first account address is selected but address is not stored in local storage
6. click on `switch wallet extension`
7. observe extension modal opens and closes immediately
8. select account from accounts dropdown
9. address gets stored in local storage
10. clicking on `switch wallet extension` still doesn't work with same behavior as on step no. 7

browser refresh doesn't help

Robiquet (approved):
Thanks, looks good. @TvrtkoM did you want to take another look?

PR #1021, opened 2023-04-13
Display real trade data + fix for sell with ztg

### LINKED ISSUE #842: Notification after successful trade should display real values, rather than pre-calculated
Parse `SwapExactAmount...` extrinsics events and display real results of a swap instead of values returned by the functions in `lib/math`

TvrtkoM:
close #842

### DIFF
--- a/components/trade-form/index.tsx
+++ b/components/trade-form/index.tsx
@@ -54,8 +54,8 @@ const getTradeValuesFromExtrinsicResult = (
     }
   }
   return {
-    baseAmount: baseAsset.div(ZTG).toFixed(4),
-    assetAmount: outcome.div(ZTG).toFixed(4),
+    baseAmount: baseAsset.div(ZTG).toFixed(1),
+    assetAmount: outcome.div(ZTG).toFixed(1),
   };
 };
 
--- a/components/trade-form/index.tsx
+++ b/components/trade-form/index.tsx
@@ -12,6 +12,7 @@ import {
   useTradeMaxBaseAmount,
   useTradeTransaction,
 } from "lib/hooks/trade";
+import { ISubmittableResult } from "@polkadot/types/types";
 import { useNotifications } from "lib/state/notifications";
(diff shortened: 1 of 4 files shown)


components/trade-form/index.tsx
```diff
@@ -27,6 +28,36 @@ import { useExtrinsic } from "lib/hooks/useExtrinsic";
...
+        baseAsset = baseAsset.add(data[0]["assetAmountIn"].toPrimitive());
+        outcome = outcome.add(data[0]["assetAmountOut"].toPrimitive());
+      } else {
+        baseAsset = baseAsset.add(data[0]["assetAmountOut"].toPrimitive());
+        outcome = outcome.add(data[0]["assetAmountIn"].toPrimitive());
+      }
+    }
+  }
+  return {
+    baseAmount: baseAsset.div(ZTG).toFixed(4),
```
Robiquet, line 57: ZTG is only worth $0.1 so 4 decimal places of precision is a bit excessive, probably 1 is plenty.

Robiquet (commented):
Looks good, is it possible to get the same values into `TradeResult`?

components/trade-form/index.tsx
TvrtkoM: they're in TradeResult - https://github.com/zeitgeistpm/ui/blob/b0352999a15222678e35232e78e754500e0ae688/components/trade-form/index.tsx#L183

TvrtkoM:
> Looks good, is it possible to get the same values into `TradeResult`?

They're in TradeResult

https://github.com/zeitgeistpm/ui/blob/b0352999a15222678e35232e78e754500e0ae688/components/trade-form/index.tsx#L183

https://github.com/zeitgeistpm/ui/blob/b0352999a15222678e35232e78e754500e0ae688/components/trade-form/index.tsx#L434

Robiquet (approved):
Oh nice, I didn't notice

Issue #960, opened 2023-04-03
Add market history modal


maltekliemann:
I can't see who made the dispute, and the time bar show 100%.

<img width="1920" alt="Screenshot 2023-04-03 at 11 55 41" src="https://user-images.githubusercontent.com/49552287/229476672-2b96fdd0-8f01-4bf4-b318-5e031d11f89d.png">

robhyrk:
Do we need a design for this?

To keep it simple we can probably just add text: "Market outcome: Disputed by `user`"

maltekliemann:
I think the usual subscan/etc. links for the user (maybe plus avatar?) and the exact time/block of the dispute are perfectly fine.

Robiquet:
Let's wait for a design, I think @mk26071980 is working on something related and there are a few other things to consider

mk26071980:
As discussed with Tom 

<img width="1003" alt="image" src="https://user-images.githubusercontent.com/80706590/231430321-c646e0c9-c6b1-4252-b885-3312cc67a365.png">
<img width="962" alt="image" src="https://user-images.githubusercontent.com/80706590/231430383-e3699c1c-3d7d-45f4-9a49-d9b0fa569572.png">

Robiquet:
> As discussed with Tom
> 
> <img alt="image" width="1003" src="https://user-images.githubusercontent.com/80706590/231430321-c646e0c9-c6b1-4252-b885-3312cc67a365.png">
> 
> <img alt="image" width="962" src="https://user-images.githubusercontent.com/80706590/231430383-e3699c1c-3d7d-45f4-9a49-d9b0fa569572.png">

@maltekliemann do you need dispute date too, or does this work for you?

robhyrk:
There is missing dispute data for `SimpleDisputes` so exploring as to why that is. Will incorporate these features for `Authorized` disputes in the meantime.

maltekliemann:
> There is missing dispute data for `SimpleDisputes` so exploring as to why that is. Will incorporate these features for `Authorized` disputes in the meantime.

Yo, `simple-disputes` is deprecated at this point. No need to look into this further, I think.

Another idea for the design: How about a complete "market history"? Something like:

> - 2023-02-03T16:12Z (block 1234567) Market opened
> - 2023-03-31T23:59Z (block 1255555) Market closed
>  - 2023-04-01T00:12Z (block 1255556) oracle {account1} reported {outcome1}
[...]

PR #1022, opened 2023-04-14
Show Asset Status in Table

robhyrk:
- [ ] add outcome labels for disputed, reported, and resolved


Robiquet:
I don't think we're going ahead with this design, have you seen this issue? https://github.com/zeitgeistpm/ui/issues/960

PR #1057, opened 2023-04-18
Disputed markets

robhyrk:
- [x] get disputed market name
- [x] use disputer info to generate user icon


robhyrk:
Proposing two options:
- use the newly created `useGetCurrentPrediction` hook to get current prediction. This hook can be modified to get all asset prices and names as well and therefore could remove `useMarketSpotPrices`
- use the existing `getCurrentPrediction` hook by passing through values from `useMarketSpotPrices` 

It seems to me that the first option involves fewer steps and could replace instances throughout the app where multiple steps are required to get the current price/prediction or assets/prices.

components/markets/MarketAssetDetails.tsx
```diff
@@ -159,11 +158,10 @@ const MarketAssetDetails = observer(({ marketId }: { marketId: number }) => {
   };
 
   const getDisputedCategoricalOutcome = () => {
+    if (!rpcMarket) return;
```
Robiquet, line 161: Is this needed? Doesn't look like the function uses it

components/markets/MarketHeader.tsx
```diff
@@ -24,32 +33,100 @@ const Tag: FC<PropsWithChildren<{ className?: string }>> = ({
...
+        <span>{status} Outome: </span>
+        <span className="font-bold">{outcome}</span>
+      </div>
+      {status !== "Resolved" && (
+        <div className="flex items-center gap-4">
+          <span>{status} by: </span>
+          <div className="flex items-center">
+            <Avatar address={by} />
+            <span className="font-medium px-3.5 text-sms h-full leading-[40px]">
+              {shortenAddress(by, 6, 4)}
```
Robiquet, line 59: Might be good to use on chain identity if one is available

components/markets/MarketHeader.tsx
```diff
@@ -24,32 +33,100 @@ const Tag: FC<PropsWithChildren<{ className?: string }>> = ({
   );
 };
 
+const MarketOutcome: FC<
+  PropsWithChildren<{ status: string; outcome: string | number; by?: string }>
```
Robiquet, line 37: There should be a `MarketStatus` type from the sdk

pages/markets/[marketid].tsx
```diff
@@ -137,17 +131,9 @@ const Market: NextPage<{
       return <NotFoundPage backText="Back To Markets" backLink="/" />;
     }
 
-    //required to fix title element warning
-    const question = indexedMarket.question;
```
Comment (Robiquet) [inline pages/markets/[marketid].tsx:141]: Did you verify that there is no longer an error after removing this?

lib/util/market-status-details.ts
```diff
@@ -0,0 +1,78 @@
+import { ZTG } from "lib/constants";
+import Decimal from "decimal.js";
+import type { ScalarRangeType } from "@zeitgeistpm/sdk/dist/types";
+import moment from "moment";
+
+export const getScalarOutcome = (
+  outcome: string,
+  scalarType: ScalarRangeType,
+) => {
+  const dateFormat = "MM.DD.YYYY";
```
Robiquet, line 10: What's the reason for the specific formatting here? I would rather format based on locale by using `Intl.formatDate`

components/markets/MarketAssetDetails.tsx
```diff
@@ -197,7 +195,8 @@ const MarketAssetDetails = observer(({ marketId }: { marketId: number }) => {
 
   return (
     <div>
-      {market?.status === "Disputed" && authReportNumberOrId != null && (
+      {/* TODO: check if this can be removed */}
```
Robiquet, line 198: If all of this information is displayed elsewhere now I think it can go

lib/gql/markets.ts
```diff
@@ -32,8 +32,26 @@ const marketQuery = gql`
       creator
       oracle
       disputeMechanism
+      resolvedOutcome
```
Robiquet, line 35: Problem with this approach is that this data is only refetched periodically every 10 minutes. So there could be some inconsistencies on the page when the status changes. Might be worth checking if we're already fetching this data in `useMarket`

pages/markets/[marketid].tsx
Comment (robhyrk) [reply pages/markets/[marketid].tsx:141]: I moved this directly into the alt tag property and it doesn't seem to show an error in the console.

components/markets/MarketAssetDetails.tsx
robhyrk: Removed. I was doing some testing as I noticed similar functions had it.

lib/gql/markets.ts
robhyrk: With that in mind; would it make more sense to get as much data as possible from `useMarket`? Data regarding reported outcomes is in there too (just seems to be missing disputed market data).

Robiquet: Yeah I think to an extent, at the moment `useMarket` is restricted by what the sdk has configured for the request. I've been thinking about this but not sure if it's best to fetch the whole market schema or just a select part of it to reduce data transfer.

robhyrk:
I've refactored to use more recent data coming from hooks versus indexedMarket as well as updating some types.

lib/gql/markets.ts
```diff
@@ -32,8 +32,25 @@ const marketQuery = gql`
       creator
       oracle
       disputeMechanism
+      report {
```
Robiquet, line 35: Can some of this be removed now?

pages/markets/[marketid].tsx
```diff
@@ -136,17 +137,16 @@ const Market: NextPage<{
...
+    useEffect(() => {
+      if (disputes && marketSdkv2.status === "Disputed") {
+        const lastDispute = disputes?.[disputes.length - 1].toHuman();
```
Comment (Robiquet) [inline pages/markets/[marketid].tsx:142]: It's best to avoid using `toHuman` if possible as it reduces type safety. Take a look at `getDisputedCategoricalOutcome` to see how you can use stuff like `isCategorical`, `asCategorical`, etc

pages/markets/[marketid].tsx
```diff
@@ -156,35 +156,22 @@ const Market: NextPage<{
...
           />
           <MarketHeader
-            question={question}
-            status={indexedMarket.status}
-            tags={indexedMarket.tags}
-            starts={starts}
-            ends={ends}
+            market={indexedMarket}
+            resolvedOutcome={marketSdkv2?.resolvedOutcome}
+            report={marketSdkv2?.report as Report}
```
Comment (Robiquet) [inline pages/markets/[marketid].tsx:167]: Any way to avoid the casting here? Maybe there is a type we can pull from the sdk

Robiquet (commented):
Thanks for the updates, looking good

pages/markets/[marketid].tsx
Comment (robhyrk) [reply pages/markets/[marketid].tsx:142]: Thanks, I took a look and did something similar and ended up using `toJSON` as this gave me access to the whole disputes object. Does that work or will it cause issues similar as `toHuman`.

pages/markets/[marketid].tsx
Comment (robhyrk) [reply pages/markets/[marketid].tsx]: I found a type called `MarketReport` that would work but it's in here `@zeitgeistpm/indexer/src/graphql/sdk` and that was causing a build error so I tweaked a few things inside `useEffect` to make it work

pages/markets/[marketid].tsx
Comment (Robiquet) [reply pages/markets/[marketid].tsx]: Yeah `toJSON` can do weird stuff too, from what I remember `toJSON` often encodes strings in a strange way and `toHuman` comma separates numbers

Robiquet (approved):

PR #1069, opened 2023-04-18
Add ability to collapse table columns if there is overflow

### LINKED ISSUE #730: Remove less important table columns on mobile
Currently market card becomes scrollable when it cannot fit all of the content, like so:
image
However, this can cover up important columns like the buy/sell buttons. We should hide less important columns (like 24hr change) to prevent this from happening. Could look like this for example:

image

Robiquet:
closes #730

### DIFF
--- a/components/assets/AssetActionButtons/TradeButton.tsx
+++ b/components/assets/AssetActionButtons/TradeButton.tsx
@@ -16,7 +16,7 @@ const TradeButton = ({
   return (
     <>
       <button
-        className="text-mariner font-semibold text-ztg-14-120"
+        className="text-mariner font-semibold text-ztg-14-120 text-center w-full"
         onClick={() => {
           setTradeItem({
             assetId: assetId,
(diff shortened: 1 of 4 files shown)


TvrtkoM (changes requested):
see comment

Robiquet:
> The columns get collapsed, but when the screen gets larger again they don't show up again. This could happen when user turns his phone in landscape position.

The problem with the current solution is that it relies on the table being rendered in order to detect if there is overflow. If we continue to check for overflow after a column has been collapsed it will create an infinite loop with the column being removed and added continuously. So for this reason once a column has been removed it cannot be added back.

Another solution would be to calculate the column width without rendering or explicitly set column width, though I think both of these solutions are much more complex.

We could also provide breakpoints with each column to indicate the screen size they should be displayed at, though I think in this case it will have the same effect as the current solution, as a landscape device probably won't hit the `sm` breakpoint.

Overall, I think that the scenario you mentioned isn't that detrimental to the user as the collapsed columns are fairly unimportant so I think the current solution is appropriate. Let me know if you think there are any other options we should consider.

TvrtkoM (approved):

PR #1071, opened 2023-04-18
Remove gql client from store and use sdkv2 indexer where it was used.

### LINKED ISSUE #1033: Refactor reliance on `store.graphQLClient` to `useSdkv2()` & `sdk.indexer.client`
Get the indexer gql client from sdk v2 instead of sdkv1 on store.

yornaath:
Fixes #1033

### DIFF
--- a/lib/constants/index.ts
+++ b/lib/constants/index.ts
@@ -91,7 +91,6 @@ const getGraphQlEndpoint = (): string => {
 };
 
 export const graphQlEndpoint = getGraphQlEndpoint();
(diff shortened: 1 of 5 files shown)


TvrtkoM (changes requested):
I added `graphQlClient` constant in `lib/constants/index.ts`. Could you remove that also and fix places where it is used?

yornaath:
> I added `graphQlClient` constant in `lib/constants/index.ts`. Could you remove that also and fix places where it is used?

done

TvrtkoM (approved):

PR #1090, opened 2023-04-20
Fix initial wallet enabling

yornaath:
When the page loads quickly(when its buildt using `yarn build && yarn start` or in production) the wallet isnt detected quickly enough so its not connected properly. This fix makes the `enable` function on the wallet poll for extension detection so it enables wallet correctly on quick page loads.

@TvrtkoM This was probably why you did the enable wallet loop on the original store?

**Reproduce error:**
- go to https://test.staging.zeitgeist.pm/
- Refresh the page multiple times.
- See that sometimes the wallet connects and sometimes its left in the `Connect Wallet` state


TvrtkoM:
> @TvrtkoM This was probably why you did the enable wallet loop on the original store?

No, I don't think that's was the case. It had something to do with avoiding refreshing the page after initial accounts are added or when the wallet is not initially given permission to cooperate with the app

TvrtkoM (approved):

PR #1097, opened 2023-04-21
Prevent body width changing on mobile menu open

### LINKED ISSUE #1056: Body of page goes full width upon clicking mobile menu

robhyrk:
closes #1056

### DIFF
--- a/components/menu/index.tsx
+++ b/components/menu/index.tsx
@@ -44,13 +44,12 @@ const TopBar = observer(() => {
   }, [changeNavBG]);
 
   useEffect(() => {
-    const scrollY = window.scrollY;
-    setScrollPosition(scrollY);
     if (menuOpen) {
-      document.body.style.position = "fixed";
+      document.body.style.height = `100%`;
+      document.body.style.overflow = `hidden`;
     } else {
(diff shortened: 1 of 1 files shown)


TvrtkoM (changes requested):
I'm not sure is it just me but I can't see or use mobile menu
This is what I see on both Brave and Chrome browsers
image

If I remove bg-white of the mobile dropdown menu element that's overlaying the top-bar the buttons are still not clickable, I can't click on any item in top bar.

I see this on both PC, Ubuntu OS and macbook, so think it's not specific to my environment. Can this be fixed with this PR?

TvrtkoM:
> I'm not sure is it just me but I can't see or use mobile menu This is what I see on both Brave and Chrome browsers image
> 
> If I remove bg-white of the mobile dropdown menu element that's overlaying the top-bar the buttons are still not clickable, I can't click on any item in top bar.
> 
> I see this on both PC, Ubuntu OS and macbook, so think it's not specific to my environment. Can this be fixed with this PR?

I'll open separate issue for this since it seems to happen in some rare cases

TvrtkoM (approved):

PR #1109, opened 2023-04-25
Use atom for storing chain time(blocknumber and timestamp)

### LINKED ISSUE #1031: Pull out value `blockTimestamp` from store to react-query hook
Create a hook that uses react-query to fetch the current block time stamp instead of reacting to the value `blockTimestamp` on Store.

yornaath:
Fixes #1031


lib/state/chaintime.ts
```diff
@@ -0,0 +1,26 @@
...
+import { sdkAtom } from "lib/hooks/useSdkv2";
+import { Observable, Subscription } from "rxjs";
+
+export const chainTimeAtom = atom<ChainTime | false>(false);
+
+const store = getDefaultStore();
+
+let sub: Subscription;
+
+store.sub(sdkAtom, () => {
```
Robiquet, line 13: Is this going to cause any "outside of react" errors?

lib/state/chaintime.ts
```diff
@@ -0,0 +1,26 @@
...
+  const sdk = store.get(sdkAtom);
+  if (sub) sub.unsubscribe();
```
Robiquet, line 15: Why do we want to unsubscribe if the subscription already exists?

lib/state/chaintime.ts
```diff
@@ -0,0 +1,26 @@
+import { isRpcSdk } from "@zeitgeistpm/sdk-next";
+import { ChainTime } from "@zeitgeistpm/utility/dist/time";
+import { atom, getDefaultStore, useAtom } from "jotai";
+import { sdkAtom } from "lib/hooks/useSdkv2";
+import { Observable, Subscription } from "rxjs";
+
+export const chainTimeAtom = atom<ChainTime | false>(false);
```
Robiquet, line 7: Why initialise as `false` instead `null` or `undefined`?

yornaath: Its a limitation of jotai/typescript. Null doesnt work somehow. Its in the jotai docs.

lib/state/chaintime.ts
yornaath: Its in case we connect a new instance of the sdk to another endpoint.

Robiquet (approved):

PR #1122, opened 2023-04-27
Mobile market filters

TvrtkoM:


yornaath:
The arrow here should maybe be aligned vertically with the text?
Screenshot 2023-05-03 at 10 12 46

components/markets/market-filter/MarketFiltersSort.tsx
```diff
@@ -0,0 +1,118 @@
...
+    return {
+      ...provided,
+      backgroundColor: "white",
+      color: "black",
+      zIndex: 100,
+    };
+  },
+};
+
+const IndicatorSeparator = () => {
```
Robiquet, line 60: Does this do anything?

components/markets/market-filter/MarketFiltersSort.tsx
```diff
@@ -0,0 +1,118 @@
...
+  }) => {
+    return (
+      <ReactSelect
+        value={marketsOrderByOptions.find((opt) => opt.value === ordering)}
+        onChange={(v) => {
+          onOrderingChange(v.value);
+        }}
+        options={marketsOrderByOptions}
+        styles={{ ...sortBySelectStyles, ...(styles ?? {}) }}
+        // menuIsOpen={true}
```
Robiquet, line 90: Can this be removed?

Robiquet (commented):
Looks nice, only think I noticed was the drop down arrow isn't clickable. Is that an easy fix?
image

components/markets/market-filter/MarketFiltersSort.tsx
TvrtkoM: Yes, it removes vertical line between select value and indicator arrow that react select is showing by default. It is added in `components` prop of ReactSelect

Robiquet (approved):

PR #1124, opened 2023-04-27
image generation - abort fetch request after 10s

TvrtkoM:
@Robiquet Please see does this help with vercel errors

### DIFF
--- a/pages/api/og/generate.tsx
+++ b/pages/api/og/generate.tsx
@@ -22,9 +22,11 @@ const getImageUrl = async (image: string | null): Promise<string> => {
   }
   const controller = new AbortController();
 
+  const abortTimeout = 10000;
+
   setTimeout(() => {
     controller.abort();
-  }, 2500);
+  }, abortTimeout);
 
   try {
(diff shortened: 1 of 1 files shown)


Robiquet:
Yeah I think this should help. Is 2.5s a reasonable amount of time to fetch an image from IPFS? I wouldn't mind it being higher if it would make more requests successful

TvrtkoM:
> Yeah I think this should help. Is 2.5s a reasonable amount of time to fetch an image from IPFS? I wouldn't mind it being higher if it would make more requests successful

Think It can be adjusted to bigger values -something like 10 seconds should work

Robiquet:
> > Yeah I think this should help. Is 2.5s a reasonable amount of time to fetch an image from IPFS? I wouldn't mind it being higher if it would make more requests successful
> 
> Think It can be adjusted to bigger values -something like 10 seconds should work

Sounds good, it will still cut 2/3 off the execution time for failing ones

Robiquet (approved):

PR #1126, opened 2023-04-27
Market history

### LINKED ISSUE #960: Add market history modal

robhyrk:
- [x]  sort events by block number in chronological order
- [x] convert block numbers into timestamps

closes #960 and #1072


lib/hooks/queries/useMarketEventHistory.ts
```diff
@@ -0,0 +1,121 @@
...
+export const useMarketEventHistory = (
+  marketId: string,
+): UseQueryResult<MarketHistory> => {
+  const [sdk2, id] = useSdkv2();
+
+  return useQuery(
+    [marketsEventsRootQuery, id, marketId],
+    async () => {
+      if (isIndexedSdk(sdk2) && isRpcSdk(sdk2)) {
+        const sdk = await getSdk();
```
yornaath, line 52: You can use the sdk2 here, no need to create a new one

lib/hooks/queries/useMarketEventHistory.ts
```diff
@@ -0,0 +1,121 @@
...
+        let { api } = sdk;
+
+        const market = await getMarket(sdk2.indexer.client, marketId);
```
yornaath, line 55: You probably want to use the useMarket hook outside of the query fetching fn

lib/hooks/queries/useMarketEventHistory.ts
robhyrk: There may be an issue with sdk2 then because the same functions for getting the timestamp return different results:

```
const blockHash = await sdk2.api.rpc.chain.getBlockHash(blockNumber);
const timestamp = (await sdk2.api.query.timestamp.now.at(blockHash)).toNumber();
```

sdk2 returns `Jul 25, 2022, 9:46 AM` for `block: 1762884`
sdk returns `Nov 24, 2022, 12:49 PM` for  `block: 1762884`

For market `423` for example, it would appear that what `sdk `returns is correct since `sdk2` returns a date that is before the market creation date.

Robiquet (changes requested):
1. Looks like numbers are chopped 
2. Report is missing timestamp
3. Modal should close when the user clicks outside 

image

Robiquet (changes requested):
1. Looks like numbers are chopped 
2. Report is missing timestamp
3. Modal should close when the user clicks outside 

image

robhyrk:
Note: markets pre migration may show incorrect open and close timestamps. Block numbers for current markets might be missing; I am opening an issue to address this on subsquid. Ideally the `historicalMarkets` query would contain data regarding outcomes, timestamps, users, and block numbers and would be updated after each event and not just after the market resolves. And/or the `markets `query should contain this sort of historical market data as well.

lib/hooks/queries/useMarketEventHistory.ts
```diff
@@ -0,0 +1,136 @@
...
+            sdk.indexer.client,
+            Number(marketId),
+          );
+          resolutionTimestamp = new Date(timestamp);
+          resolutionBlock = blockNumber;
+        }
+
+        const getTimeStampForBlock = async (blockNumber: number) => {
+          try {
+            const blockHash = await getApiAtBlock(sdk.api, blockNumber);
```
Robiquet, line 83: `blockHash` is of type `ApiPromise` so the naming is a bit misleading

lib/hooks/queries/useMarketEventHistory.ts
```diff
@@ -0,0 +1,136 @@
...
+            return await blockHash.query.timestamp
+              .now()
+              .then((now) => now.toNumber());
```
Robiquet, line 86: Think this line is redundant as you are already awaiting the promise

Robiquet (commented):
Thanks for the updates, just a few small things

Robiquet (approved):
Looks good, thanks!

PR #1127, opened 2023-04-28
Retain market list scroll

### LINKED ISSUE #1009: Retain scroll position in market list
Scroll position in market list should be retain when navigating between a market page and the market list.

Refactoring is required so that query params are always loaded upfront

### LINKED ISSUE #1098: Index page scroll position not retained
1. Click a market in "Trending Markets"
2. Navigate back from market page

Scroll position moves back to the top of the page. Instead it should go back to "Trending Markets"

Robiquet:
close #1009 close #1098

### DIFF
--- a/components/markets/MarketsList.tsx
+++ b/components/markets/MarketsList.tsx
@@ -62,6 +62,12 @@ const MarketsList = observer(({ className = "" }: MarketsListProps) => {
 
   const { ref: loadMoreRef, inView: isLoadMarkerInView } = useInView();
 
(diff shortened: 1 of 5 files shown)


yornaath (changes requested):
Scroll doesnt seem to be retained properly when I test the preview.

Robiquet:
> Scroll doesnt seem to be retained properly when I test the preview.

Seems ok to me on Brave, could you give some steps to reproduce?

yornaath (approved):
Works!

PR #1138, opened 2023-05-02
Show block reason. Refactor transaction button

Robiquet:


components/ui/FormTransactionButton.tsx
```diff
@@ -1,51 +1,27 @@
...
+  children,
+}) => {
+  return (
+    <TransactionButton
+      type="submit"
+      disabled={disabled}
+      className={className}
+      dataTest={dataTest}
+      children={children}
+    ></TransactionButton>
```
yornaath, line 23: I think this is the more "reacty" way of passing children as opposed to using a prop?

```suggestion
    <TransactionButton
      type="submit"
      disabled={disabled}
      className={className}
      dataTest={dataTest}
    >
      {children}
    </TransactionButton>
```

yornaath (changes requested):
Just a nitpicky change request, but looks good

yornaath (approved):

PR #1157, opened 2023-05-05
Scalar util fn for displaying outcomes

robhyrk:
found myself re-using this function so created one that can be shared throughout the app


lib/util/get-scalar-outcome.ts
```diff
@@ -0,0 +1,22 @@
+import { Decimal } from "decimal.js";
+import { ZTG } from "../constants";
+import type { ScalarRangeType } from "@zeitgeistpm/sdk/dist/types";
+
+export const getScalarOutcome = (
```
yornaath, line 5: I would rename this to `formatScalarOutcome` since its not strictly a getter of the outcome but rather a display formatter.

components/markets/ScalarPriceRange.tsx
```diff
@@ -28,44 +29,20 @@ const ScalarPriceRange = ({
   const inferedType: ScalarRangeType = scalarType ?? "number";
```
yornaath, line 29: Do you need to do a null check here since its also done and inferred in the formatting function?

yornaath (changes requested):
Looks good, just a couple small changes requested

Robiquet (approved):

yornaath (approved):

PR #1169, opened 2023-05-09
fix reporting

### LINKED ISSUE #1168: After market closes, reporting buttons doesn't show up
Steps to reproduce:

1. Create a market on app UI so that it will end quickly, with no grace period and 1hr oracle duration deadline
2. Wait for market to close
3. Notice timeline says that market ended but oracle is not able to report since report buttons doesn't show up

TvrtkoM:
close #1168


Robiquet (approved):
Thanks, good spot

PR #1174, opened 2023-05-10
replace liquidity hook

### LINKED ISSUE #1170: For Liquidity pools list, Reuse hook used for markets page
Now that we have pools data under markets schema, the same hook used on markets page can be used on the liquidity page to avoid complications when filtering pools based on market data such as market statuses

TvrtkoM:
close #1170


lib/hooks/queries/useTotalLiquidity.ts
```diff
@@ -1,42 +1,40 @@
 import { useQuery } from "@tanstack/react-query";
 import Decimal from "decimal.js";
 import { useSdkv2 } from "../useSdkv2";
+import { ZTG } from "@zeitgeistpm/sdk-next";
 
 export const rootKey = "total-liquidity";
 
-export const useTotalLiquidity = (options: { enabled: boolean }) => {
+const calcLiqudity = (
+  assets: { price: string | number; amountInPool: any }[],
```
Robiquet, line 9: Could you add a type for `amountInPool`?

lib/hooks/queries/useTotalLiquidity.ts
```diff
@@ -1,42 +1,40 @@
...
-  );
-
-  const { data: saturatedIndex } = useQuery(
-    [id, rootKey, "saturated-index"],
     async () => {
-      return sdk.model.swaps.saturatedPoolsIndex(pools);
+      const pools = await sdk.model.swaps.listPools({});
+      const total =
+        pools?.reduce((acc, pool) => {
+          return acc.plus(calcLiqudity(pool.assets as any));
```
Robiquet, line 31: This won't be quite right now that we have different base assets in the pool. We will need to convert those to be denominated in ZTG. Happy to move this to another ticket though as it was already like this

pages/liquidity/index.tsx
```diff
@@ -1,19 +1,40 @@
...
 import { useTotalLiquidity } from "lib/hooks/queries/useTotalLiquidity";
 import { useZtgPrice } from "lib/hooks/queries/useZtgPrice";
+import { MarketsOrderBy } from "lib/types/market-filter";
 import { formatNumberLocalized } from "lib/util";
 import { NextPage } from "next";
 import { useRouter } from "next/router";
 import { useMemo } from "react";
 import { AiOutlineRead } from "react-icons/ai";
 
+const calcLiqudity = (
```
Robiquet, line 24: Looks like we have 2 `calcLiqudity` functions, are they both needed?

pages/liquidity/index.tsx
```diff
@@ -41,74 +62,87 @@ const columns: TableColumn[] = [
...
-            {
-              value: saturatedData?.liquidity.div(ZTG).toNumber(),
-              usdValue: ztgPrice?.toNumber() ?? 0,
-            }
-          ) : (
-            <span>...</span>
-          ),
+          poolBalance: {
+            value: poolLiquidty.toNumber(),
+            usdValue: ztgPrice?.toNumber() ?? 0,
```
Robiquet, line 140: This should be `ztgPrice.mul(poolLiquidty)`

PR #1218, opened 2023-05-15
Remove liquidity page

### LINKED ISSUE #1204: Temporarily remove the liquidity pools list page
Maybe not temporarily

TvrtkoM:
close #1204


Robiquet:
@lsaether what's the reason for this? We recently updated it and it's working pretty well now

lsaether:
> @lsaether what's the reason for this? We recently updated it and it's working pretty well now

The feedback we got was that it's confusing to users, and basically presents the markets in a different view that is less useful than the markets page. Since there's liquidity details on the markets now, there's no extra information given from the liquidity page. So, we're going to workshop the liquidity page and in the meantime remove it to get it out of the way.

PR #1226, opened 2023-05-16
Form validation fixes

### LINKED ISSUE #1156: Sell full set max limit is too large
1. go to liquidity page of any active market
2. you should have some full sets and if not. buy some on `buy full set` modal
3. now open `sell full set` modal and observe form allows you to sell more full sets than you have

### LINKED ISSUE #1164: Pool can only be drained completely once it is resolved
Currently form validation for exit pool only checks if market is active to set form invalid if user tries to take everything out from pool. It is impossible to do so until market is resolved.

TvrtkoM:
close #1156 
close #1164

### DIFF
--- a/components/liquidity/ExitPoolForm.tsx
+++ b/components/liquidity/ExitPoolForm.tsx
@@ -237,7 +237,7 @@ const ExitPoolForm = ({
                       market.status.toLowerCase() !== "resolved" &&
                       poolAssetBalance.minus(value).lessThanOrEqualTo(0.01)
                     ) {
-                      return "Pool cannot be emptied completely before market gets resolved";
+                      return "Pool cannot be emptied completely before the market resolves";
                     }
                   },
                 })}


components/liquidity/ExitPoolForm.tsx
```diff
@@ -236,10 +234,10 @@ const ExitPoolForm = ({
                     } else if (value <= 0) {
                       return "Value cannot be zero or less";
                     } else if (
-                      poolStatus.toLowerCase() === "active" &&
+                      market.status.toLowerCase() !== "resolved" &&
```
Robiquet, line 237: Are you sure this is correct? I've discussed this with Malte previously, I can double check though

Robiquet: Checked with him, what you've changed is correct

components/liquidity/ExitPoolForm.tsx
```diff
@@ -236,10 +234,10 @@ const ExitPoolForm = ({
...
                       poolAssetBalance.minus(value).lessThanOrEqualTo(0.01)
                     ) {
-                      return "Pool cannot be emptied completely whilst it's active";
+                      return "Pool cannot be emptied completely before market gets resolved";
```
Robiquet, line 240: Maybe this is slightly better?
```suggestion
                      return "Pool cannot be emptied completely before the market resolves";
```

Robiquet (approved):

PR #1237, opened 2023-05-16
Portfolio value denomination

### LINKED ISSUE #1114: [Multi-asset] Portfolio summary denomination
Does it still make sense to denominate in ZTG if we have multiple base assets? 
image

Robiquet:
closes #1114 

Also fixes usd values on market page and popular categories link

### DIFF
--- a/components/portfolio/BondsTable.tsx
+++ b/components/portfolio/BondsTable.tsx
@@ -1,13 +1,10 @@
-import { IOForeignAssetId, parseAssetId } from "@zeitgeistpm/sdk-next";
-import Table, { TableColumn, TableData } from "components/ui/Table";
+import Table, { TableColumn } from "components/ui/Table";
 import Decimal from "decimal.js";
 import { ZTG } from "lib/constants";
 import { useAccountBonds } from "lib/hooks/queries/useAccountBonds";
-import {
-  ForeignAssetPrices,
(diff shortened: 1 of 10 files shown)


components/portfolio/BondsTable.tsx
```diff
@@ -28,8 +34,22 @@ const columns: TableColumn[] = [
   },
 ];
 
+const lookUpBaseAssetPrice = (
```
TvrtkoM, line 37: code in this function is used in other places and possibly more in the future. Can it be moved to `util` somewhere and so it can be reused?

components/portfolio/MarketPositions.tsx
```diff
@@ -78,17 +82,33 @@ export const MarketPositions = ({
...
-          ({ assetId, price, userBalance, outcome, changePercentage }) => {
+          ({
+            assetId,
+            price,
+            userBalance,
+            outcome,
+            changePercentage,
+            market,
+          }) => {
+            const baseAssetId = parseAssetId(market.baseAsset).unrightOr(null);
```
TvrtkoM, line 93: those 4 lines e.g. can they also be replaced with `lookupBaseAssetPrice` function?

TvrtkoM (changes requested):
See comments made on code

TvrtkoM (approved):

PR #1241, opened 2023-05-17
Query params pages states

### LINKED ISSUE #1165: Persist pages tab selection in query parameters
Example:
1. go to portfolio page
2. click on subsidy tab
3. navigate to a liquidity pool page
4. now navigate back

notice this is not the section you were before. It shows `my markets` instead of `subsidy`. This can be fixed by persisting the tab selection inside query parameters.

Every page that has tab selection should store it in query parameter.

TvrtkoM:
close #1165


lib/util/get-query-params.ts
```diff
@@ -0,0 +1,18 @@
+import { parse as parseUri } from "uri-js";
+
+/**
+
+Extracts query parameters from the given path and returns them as an object.
+Router from next.js does not provide a way to get only the query parameters from router, so this function is used to extract them.
+@param path - The path containing the query parameters.
+@returns An object representing the extracted query parameters.
+*/
+export const getQueryParams = (path: string): { [key: string]: any } => {
```
robhyrk, line 10: Can there be a more specific type here then `any` ?

robhyrk (commented):
Looks good to me and everything seems to function well. I just left a comment regarding a type in `get-query-params` wondering if you could use something more specific than `any`

lib/util/get-query-params.ts
robhyrk, line 10: Can there be a more specific type here then `any` ?

robhyrk (changes requested):

robhyrk (approved):
looks good to me!

PR #1267, opened 2023-05-24
Portfolio account switch

### LINKED ISSUE #1236: Changing selected account on portfolio page should reload portfolio with new account

TvrtkoM:
close #1236

### DIFF
--- a/components/markets/MarketTimer.tsx
+++ b/components/markets/MarketTimer.tsx
@@ -69,7 +69,7 @@ export const MarketTimer = ({ stage }: MarketTimerProps) => {
 
 export const MarketTimerSkeleton = () => {
   return (
-    <div className="inline-block w-full max-w-[800px]">
+    <div className="inline-block w-full">
       <div className="flex mb-4 items-center">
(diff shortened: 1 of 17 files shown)


Robiquet (commented):
Not too sure how it happened but I got this error when switching between accounts 
image

TvrtkoM:
> Not too sure how it happened but I got this error when switching between accounts image

Can't reproduce it

TvrtkoM:
> I think we should still support account sharing if there is a connected wallet. For example if I click the link on the market page I get redirected to the connected wallet isn't of the one in the link I clicked image

You're right this should work, but what happens if i change the account now?
e.g. my wallet is connected and, open Malte's portfolio and now I switch account in my wallet. Think I should still be on Malte's portfolio page

Robiquet:
> update portfolio account switching

yeah that sounds right to me

Robiquet:
> > Not too sure how it happened but I got this error when switching between accounts image
> 
> Can't reproduce it

I've pushed a fix

Robiquet (approved):

PR #1270, opened 2023-05-25
Remove fees from spot price calculation

### LINKED ISSUE #1265: "Prediction after trade" shouldn't include swap fee

Robiquet:
closes #1265 
Before:
image

After:
image

Price after trade:
image

(Prices after trade are closer when we don't count fees)


Robiquet:
> Price after arbitrage shouldn't include fees. We should show pool fees separately. Info bubbles to explain things

We are reworking the design for some of this section now so I think this PR is good to go it it's current state

TvrtkoM (approved):

PR #1275, opened 2023-05-26
Fix some strictNullCheck errors

Robiquet:


components/account/AccountButton.tsx
```diff
@@ -246,14 +248,14 @@ const AccountButton: FC<{
                           <BalanceRow
                             imgPath="/currencies/ztg.jpg"
                             units={constants?.tokenSymbol}
-                            balance={activeBalance}
+                            balance={activeBalance ?? undefined}
```
TvrtkoM, line 251: think it's better to make `activeBalance` be `Decimal | undefined` by modifying `useBalance` hook
e.g. 

```
  const query = useQuery<Decimal | undefined>(
  [...],
async() => {
....
/// no return at the end
}, {}):
```

components/liquidity/LiquidityModal.tsx
```diff
@@ -65,7 +65,9 @@ const LiquidityModal = ({
     })
     ?.data.balance.free.toString();
 
-  const baseAsset = parseAssetId(pool?.baseAsset).unrightOr(null);
+  const baseAsset = pool?.baseAsset
```
TvrtkoM, line 68: this can be written as `const baseAsset = pool && parseAssetId(pool.baseAsset).unrightOr(undefined);`

`&&` operator returns the value of the first falsy operand encountered when evaluating from left to right, or the value of the last operand if the first is truthy, so baseAsset will be undefined if pool is undefined and if pool is defined it will return the value of `parseAssetId(pool.baseAsset).unrightOr(undefined)`

lib/hooks/queries/useAccountPoolAssetBalances.ts
```diff
@@ -16,11 +22,15 @@ export const useAccountPoolAssetBalances = (
...
     async () => {
-      if (isRpcSdk(sdk)) {
+      if (isRpcSdk(sdk) && pool) {
         const assets = isIndexedData(pool)
           ? pool.weights
-              .filter((weight) => weight.assetId !== "Ztg")
-              .map((weight) => JSON.parse(weight.assetId))
+              .filter((weight) => weight?.assetId !== "Ztg")
+              .map((weight) =>
+                weight?.assetId
```
TvrtkoM, line 30: same here. Personally I'd rather use this `weight && parseAssetId(weight.assetId).unrightOr(undefined)`

TvrtkoM (approved):

PR #1293, opened 2023-05-31
fix modal styling issues

### LINKED ISSUE #1201: Mobile menu should disable scrolling
When the mobile menu is open, app shouldn't be able to scroll content in the background.

`Dialog` component from headlessui should handle this automatically

### LINKED ISSUE #1281: horizontal scrollbar on the bottom
There is horizontal scrollbar on the bottom of every page. Source of this is in the following css set on `html`

```
  overflow: scroll!important;
  padding-right: 0!important;
```

It's an issue with headless ui modals that add padding or something which makes the page jitter when there is a modal. Maybe there's another way to avoid jitterness so there is no visible scrollbar

TvrtkoM:
close #1281 
close #1201


Robiquet:
Think creates very janky behaviour when opening this modal and scrolling. I can scroll a bit but then it brings me back up
image

Robiquet (approved):

PR #1297, opened 2023-05-31
Hide dust balances within portfolio

### LINKED ISSUE #1268: Sell button should be disabled when all assets are sold
Screenshot 2023-05-24 at 9 51 48 PM

robhyrk:
closes #1268 

Solution: displays "<0.01" for balances that are greater than 0 but above 0.01 so that they can still be sold since trading is allowed for values as low as 0.001

### DIFF
--- a/components/portfolio/MarketPositions.tsx
+++ b/components/portfolio/MarketPositions.tsx
@@ -132,11 +132,6 @@ export const MarketPositions = ({
       foreignAssetPrices,
       usdZtgPrice,
     );
-    console.log(
-      pos.userBalance.mul(pos.price).mul(baseAssetUsdPrice).div(ZTG).toNumber(),
-      pos.outcome,
-      market.marketId,
-    );
     return (
       pos.userBalance
         .mul(pos.price)
--- a/components/portfolio/MarketPositions.tsx
+++ b/components/portfolio/MarketPositions.tsx
@@ -19,6 +19,7 @@ import Link from "next/link";
 import MarketPositionHeader from "./MarketPositionHeader";
 import { useAllForeignAssetUsdPrices } from "lib/hooks/queries/useAssetUsdPrice";
 import { lookUpAssetPrice } from "lib/util/lookup-price";
+import { USD_ONE_CENT } from "lib/constants";
 
 const COLUMNS: TableColumn[] = [
   {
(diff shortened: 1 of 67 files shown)


robhyrk:
Updated: display rows with total USD values above $0.01

components/portfolio/MarketPositions.tsx
```diff
@@ -125,197 +125,98 @@ export const MarketPositions = ({
...
-            rpnl,
-            upnl,
-          }) => {
-            const baseAssetUsdPrice = lookUpAssetPrice(
-              market.baseAsset,
-              foreignAssetPrices,
-              usdZtgPrice,
-            );
+  //display balance if any of the positions is greater than 0.01 USD
+  const displayBalance = positions.some((pos) => {
```
Robiquet, line 129: What happens if one of the balances in a market isn't above $0.01? Won't it still be displayed?

components/portfolio/MarketPositions.tsx
```diff
@@ -125,197 +125,98 @@ export const MarketPositions = ({
...
+      market.baseAsset,
+      foreignAssetPrices,
+      usdZtgPrice,
+    );
+    return (
+      pos.userBalance
+        .mul(pos.price)
+        .mul(baseAssetUsdPrice)
+        .div(ZTG)
+        .toNumber() >= 0.01
```
Robiquet, line 140: Could you pull 0.01 out to an appropriately named variable so that it's clearer?

Robiquet (commented):
Comment

components/portfolio/MarketPositions.tsx
robhyrk: Yes, good catch. I've made the appropriate changes in the latest push to hide rows and then entire tables if all the values are below the threshold.

Robiquet (approved):

components/portfolio/MarketPositions.tsx
```diff
@@ -19,6 +19,7 @@ import Link from "next/link";
 import MarketPositionHeader from "./MarketPositionHeader";
 import { useAllForeignAssetUsdPrices } from "lib/hooks/queries/useAssetUsdPrice";
 import { lookUpAssetPrice } from "lib/util/lookup-price";
+import { USD_ONE_CENT } from "lib/constants";
```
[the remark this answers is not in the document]
robhyrk: I like that better and made the change

PR #1306, opened 2023-06-02
Exit pool form updates

### LINKED ISSUE #1167: Inputs disappear after whole pool is drained
After market resolves and as a liquidity provider I pull everything out by using Exit pool form, after successful extrinsic all inputs disappear and only slider is left inside modal.

Better if we could change this to disabling the input for the asset that doesn't have any pool balance

TvrtkoM:
close #1167


Robiquet:
Not sure I really like this behaviour, it seems inconsistent with what we have done elsewhere. Just because the user doesn't have assets we still let them view the trade form and the join pool form.
image

Robiquet (approved):

PR #1309, opened 2023-06-05
Update useMarketEventHistory with updated historicalMarkets query

robhyrk:
- [ ] query historical markets by marketId
- [ ] update and refactor useMarketEventHistory hook

### DIFF
--- a/lib/gql/market-history.ts
+++ b/lib/gql/market-history.ts
@@ -1,0 +1,43 @@
+import { gql, GraphQLClient } from "graphql-request";
+import { MarketStatus } from "@zeitgeistpm/sdk-next";
+import { OutcomeReport } from "@zeitgeistpm/sdk/dist/types";
+
+const historicalMarketQuery = gql`
+  query HistoricalMarkets($marketId: Int) {
+    historicalMarkets(where: { marketId_eq: $marketId }) {
+      by
+      blockNumber
+      event
+      id
+      outcome {
+        categorical
+        scalar
+      }
+      resolvedOutcome
+      status
+      timestamp
+    }
+  }
+`;
+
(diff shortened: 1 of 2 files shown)


components/markets/MarketHeader.tsx
```diff
@@ -231,32 +240,33 @@ const MarketHistory: FC<
...
                       : getScalarOutcome(
-                          marketHistory?.resolved?.outcome,
+                          marketHistory?.resolved?.resolvedOutcome,
                           scalarType,
                         )}
                   </span>
                 </p>
                 <span className="text-sm text-gray-500">
-                  {marketHistory?.resolved?.timestamp &&
+                  {marketHistory?.resolved?.timestamp > 0 &&
```
Robiquet, line 264: This check seems a bit strange to me, can `timestamp` be 0 or less?

lib/gql/market-history.ts
```diff
@@ -0,0 +1,45 @@
...
+  outcome: OutcomeReport;
+  resolvedOutcome: string;
+  timestamp: number;
+};
+
+export const getMarketHistory = async (
+  client: GraphQLClient,
+  marketId: number,
+): Promise<MarketEvent[]> => {
+  if (!marketId) return [];
```
Robiquet, line 40: MarketId isn't optional so this check shouldn't be needed, or maybe the types need to be adjusted. What if marketId is 0?

components/markets/MarketHeader.tsx
robhyrk: Ah yeah, no need for that anymore with the newer iteration. I've made the change now.

Robiquet (approved):

PR #1313, opened 2023-06-05
Leaderboard

### LINKED ISSUE #37: Add Leaderboard page
Right now I think it's sufficient to just have a global rank. In time, we're going to want to calculate this for daily, weekly, monthly, yearly, and possibly other intervals. Also, we'll eventually be ranking accounts by market subcategories and by different account attributes.

Robiquet:
closes #37 

- [x] Remember to make the leaderboard link in `components/menu/index` clickable by removing preventDefaults and set correct colors + remove coming soon label. The link is defined in two places on inside the menu when on mobile and one directly on the topbar for desktop.


yornaath:
@Robiquet Made a checkpoint for remembering to make the leaderboard link active before this is merged. If pull 1358 is merged first.

TvrtkoM (changes requested):
my identity name is not displayed correctly
image

it should be `TMajstorović`

also, see comments

TvrtkoM (approved):

PR #1317, opened 2023-06-06
disable redeem button when browsing foreign account portfolio

### LINKED ISSUE #1285: Portfolio - shouldn't be able to redeem if it is not portfolio of connected wallet
If I go to someone else's portfolio I can see enabled redeem buttons for markets that are resolved. It should be either hidden or disabled.

TvrtkoM:
close #1285


components/assets/AssetActionButtons/RedeemButton.tsx
```diff
@@ -119,6 +120,11 @@ export const RedeemButtonByValue = ({
   const wallet = useWallet();
   const signer = wallet?.getActiveSigner();
   const notificationStore = useNotifications();
+  const router = useRouter();
+  const walletAddress = wallet?.selectedAddress;
+  const addressFromUrl = Array.isArray(router.query.address)
```
Robiquet, line 125: Not sure if this is the best approach, won't this cause the button to be disabled if it is used on a page that doesn't have the user's address in the route? I think it would be better to do a balance check for the connected wallet

Robiquet (approved):

PR #1330, opened 2023-06-08
Profile tooltips

### LINKED ISSUE #1274: Display tooltips with explanations for columns on portfolio page
It would be beneficial to include information explaining each column on the portfolio page to user. Below are some suggestions for the descriptions:

1. **Outcomes (Name of the Asset)**: These are the results you've purchased in the prediction market. Each different outcome is linked to a unique asset in your portfolio.

2. **Balance (Amount of Asset)**: This is the amount of each asset you currently hold in your portfolio. It's equivalent to the number of shares you have for a specific outcome in a prediction market.

3. **Price (Current Trading Price of the Asset)**: This is the current going rate for the asset in the market. This price can fluctuate depending on how other users are buying and selling the asset.

4. **Average Cost**: This is the average price you've paid for each asset in your portfolio. If you bought an asset multiple times at different prices, this number takes all those transactions into account.

5. **Total Value**: This is the current worth of your holdings for a specific asset. It's calculated by multiplying the amount of the asset you own (your balance) by the asset's current market price.

6. **Unrealized PnL (Profit and Loss)**: This is the profit or loss you would make if you were to sell your assets at the current market price. It's the difference between the current market price and the average cost of your assets, multiplied by the amount of the asset you own. Note: this amount doesn't reflect slippage or trading fees.

7. **Realized PnL (Profit and Loss)**: This is the actual profit or loss you've made from selling assets in your portfolio. It's the difference between the price you sold your assets at and the average cost of those assets, multiplied by the quantity of the asset that you sold.

8. **24 Hours (Change in Asset Price Over Last 24 Hours)**: This shows how much the price of each asset in your portfolio has changed in the last 24 hours. It's a quick way to track the recent performance of your assets and gauge short-term market trends.

robhyrk:
closes #1274 

Feel free to suggest another way to explain the info text if you have ideas. Thanks!


components/ui/Table.tsx
```diff
@@ -459,6 +464,12 @@ const Table = ({
                           ) : (
                             <></>
                           )}
+                          {column.infobox && (
+                            <InfoPopover
+                              title="Test"
```
Robiquet, line 469: Think this should be adjusted to something meaningful or removed

robhyrk: Sorry, forgot to remove this

Robiquet (approved):

PR #1334, opened 2023-06-08
add redeem info to other transactions on portfolio

### LINKED ISSUE #1324: Display reedeem info on transaction history

TvrtkoM:
close #1324


lib/util/merge-sorted-arrays.ts
```diff
@@ -0,0 +1,39 @@
+import PriorityQueue from "js-priority-queue";
+
+interface Entry<T> {
+  array: T[];
+  index: number;
+}
+
+function mergeSortedArrays<
```
Robiquet, line 8: Why can't we just merge then sort by block number after?

TvrtkoM: this is more efficient on large datasets

Robiquet: How much time does this save for a typical user? Seems a bit overkill to add a new library for this. I am a bit skeptical or whether this is actually more performant as it's still looping over both of the arrays

TvrtkoM: image

TvrtkoM: image

TvrtkoM: image

Robiquet: It takes less than 1ms to sort an array of 10000 elements so I don't think this is something we need to optimse
image

package.json
```diff
@@ -100,6 +100,7 @@
     "@testing-library/react": "^14.0.0",
     "@types/flexsearch": "^0.7.2",
     "@types/jest": "^29.5.0",
+    "@types/js-priority-queue": "^0.0.6",
```
Robiquet, line 103: Think this can be removed too

Robiquet (approved):

PR #1343, opened 2023-06-09
Eslint strict null

Robiquet:


TvrtkoM:
I tried this myself and eslint doesn't recognize this plugin. We need to agree how to migrate to strict inside team.
Think the best way to do it is to manually change strict to true in tsconfig when working in future pull requests and try to make that code without strict errors

PR #1348, opened 2023-06-10
strict liquidity components

TvrtkoM:


components/liquidity/LiquidityModal.tsx
```diff
@@ -93,24 +91,33 @@ const LiquidityModal = ({
...
             ? new Decimal(poolBaseBalance.toString())
-            : new Decimal(poolAssetBalances[index].free.toString());
+            : new Decimal(
+                poolAssetBalances == null
+                  ? 0
+                  : poolAssetBalances[index].free.toString(),
+              );
 
-          const id = assetObjStringToId(weight.assetId);
+          const id = assetObjStringToId(weight!.assetId);
```
Robiquet, line 107: I think this is ok for now but in the future I rather fix this by changing the sdk types, I don't think it makes sense for a weight to be `null`, if it doesn't exist it should be in the array

components/liquidity/MarketLiquiditySection.tsx
```diff
@@ -119,7 +121,7 @@ const LiquidityHeader = ({ market }: { market: FullMarketFragment }) => {
         </LiquidityHeaderButtonItem>
       </div>
       <LiquidityModal
-        poolId={pool.poolId}
+        poolId={pool!.poolId}
```
Robiquet, line 124: Maybe this should be fixed on the sdk too

components/liquidity/PoolTable.tsx
```diff
@@ -47,39 +47,43 @@ const PoolTable = ({
...
-        .mul(spotPrices?.get(index) ?? 0)
-        ?.mul(baseAssetUsdPrice ?? 0);
-      category = market?.categories[index];
-    }
+      if (IOBaseAssetId.is(assetId)) {
+        amount = basePoolBalance ?? undefined;
+        usdValue = basePoolBalance?.mul(baseAssetUsdPrice ?? 0);
+        category = { color: "#ffffff", name: metadata?.symbol };
+      } else {
+        amount = new Decimal(balances![index]?.free.toString() ?? 0);
```
Robiquet, line 65: Probably safer to do this
```suggestion
        amount = new Decimal(balances?.[index]?.free.toString() ?? 0);
```

lib/hooks/queries/useAccountAssetBalances.ts
```diff
@@ -14,7 +15,7 @@ export type UseAccountAssetBalances = {
   ) => UseQueryResult<
     {
       pair: AccountAssetIdPair;
-      balance?: OrmlTokensAccountData;
+      balance: OrmlTokensAccountData | undefined;
```
Robiquet, line 18: Isn't the equivalent to what was there previously?

lib/hooks/queries/useAccountAssetBalances.ts
```diff
@@ -71,36 +72,52 @@ export const useAccountAssetBalances = (
...
-            const balance = !pair.account
-              ? null
-              : await api.query.tokens.accounts(pair.account, pair.assetId);
-
-            return {
-              pair,
-              balance,
-            };
+          const api = await getApiAtBlock(
+            (sdk as RpcContext)!.api,
```
Robiquet, line 76: Think there is a function on the sdk for this `asRpc`

lib/hooks/queries/useTotalIssuanceForPools.ts
```diff
@@ -54,8 +54,8 @@ const batcher = memoize((sdk: Sdk<RpcContext>) => {
       }));
     },
     scheduler: batshit.windowScheduler(10),
-    resolver: (data, query) => {
-      return data.find((d) => d.poolId === query);
+    resolver: (data: PoolTotalIssuance[], query) => {
+      return data.find((d) => d.poolId === query) as PoolTotalIssuance;
```
Robiquet, line 58: Seems strange that casting is needed here. Something to be adjust in batshit?

lib/hooks/queries/useAccountAssetBalances.ts
```diff
@@ -71,36 +72,52 @@ export const useAccountAssetBalances = (
...
+  const get = useMemo(
+    () => (account: string, assetId: AssetId) => {
+      const query = queries.find(
+        (q) =>
+          q.data != null &&
+          q.data.pair.account === account &&
+          JSON.stringify(q.data.pair.assetId) === JSON.stringify(assetId),
+      );
+      if (query?.data === undefined) {
+        throw new Error(
```
Robiquet, line 111: Are you sure we should throw and error here? Won't that stop code execution?

components/liquidity/MarketLiquiditySection.tsx
TvrtkoM: It's ok if pool doesn't exist, but think sdk should remove `null` usage on most types if possible

lib/hooks/queries/useAccountAssetBalances.ts
TvrtkoM: This should never happen in reality though since we're iterating over given `AccountAssetIdPair[]`. If it does happen `get` is used where it should not have been used.

components/liquidity/MarketLiquiditySection.tsx
TvrtkoM: Tried adding new type to use in components in this file
```
type FullMarketFragmentWithPool = FullMarketFragment & {
  pool: NonNullable<FullMarketFragment["pool"]>;
};
```

but then code on marketId page complains. Should I add the type above and the type guard for it or using `pool!` is fine in this case since pool is required for those components to work

Robiquet (approved):

yornaath:
@TvrtkoM the changes made to PoolSettings and PoolTable are critical for the market create page so please be carefull when merging them 🙏

PR #1370, opened 2023-06-15
Updated cards

robhyrk:
- [x] update scalar markets (waiting for design)
- [x] confirm icons, categories, and. tags with design


components/markets/market-card/index.tsx
```diff
@@ -45,26 +47,63 @@ const Pill = ({ value, classes }: { value: string; classes: string }) => {
...
+  isVerified,
+}: {
+  tags: string[];
+  baseAsset: string;
+  isVerified: boolean;
+}) => {
   return (
     <>
-      {tags?.map((tag, index) => {
+      {/* update later to include other assets */}
```
Robiquet, line 78: Could you make this dynamic? Should be able to to if with `lookupAssetImagePath` or something similar

components/markets/market-card/index.tsx
```diff
@@ -45,26 +47,63 @@ const Pill = ({ value, classes }: { value: string; classes: string }) => {
...
+              height={20}
+              src="icons/politics-cat-icon.svg"
+              alt="politics"
+            />
+          )
         );
-      })}
+      })} */}
+      {isVerified && (
+        <Image width={20} height={20} src="icons/verified-icon.svg" alt="" />
```
Robiquet, line 105: Missing `alt` text

components/markets/MarketHeader.tsx
```diff
@@ -27,7 +27,7 @@ import { formatScalarOutcome } from "lib/util/format-scalar-outcome";
 import { Dialog } from "@headlessui/react";
 import { usePoolLiquidity } from "lib/hooks/queries/usePoolLiquidity";
 
-const UserIdentity: FC<
+export const UserIdentity: FC<
```
Robiquet, line 30: Could you move this to a separate file?

Robiquet (commented):
Not too sure what's going on here? Prediction is 0 but the bar is in the middle
image

Robiquet (commented):
Not too sure what's going on here? Prediction is 0 but the bar is in the middle
image

Robiquet:
Could you remove the image fetching from IPFS? Seems like they need to be driven from the CMS

lsaether:
Could we drop the word "Current" from the scalar bar so it just reads "Prediction:"?

lsaether:
When you hover over the market creator you see the "+" interaction, but clicking on this still brings you to the market page. Do we need to define what the behavior should be here?

robhyrk:
> When you hover over the market creator you see the "+" interaction, but clicking on this still brings you to the market page. Do we need to define what the behavior should be here?

This happens because the entire card is clickable to the market so clicking on anything within the card will cause that. I'm able to disable the avatar being clickable so I've done that in the latest update. 

The "+" that appears when you hover over it is supposed to copy the creator's address. I'm unsure if a user would find that useful on the cards, but we could explore keeping it or adding a different behaviour.  One way to make the avatar clickable is to make the top portion of cards un-clickable, thus making the creator avatar clickable.

robhyrk:
Addressed issues mentioned and added scalar lower and upper bounds back discreetly. We could consider removing the bounds on the market cards and only keeping them on the market page as an alternative.

Robiquet:
We might need to swap this image out as the background is showing up 
image

lsaether:
@robhyrk Ok, I think that disabling the copy interaction on the avatar on the card makes sense!

Robiquet (approved):

PR #1401, opened 2023-06-20
Style tooltips on create page and tables according to designs

yornaath:
Notices that the current info popovers didnt look like the designs in the create page designs so restyled them.
Also noticed that a copy paste of the <InfoPopover /> component was used in the table so consolidated into one component.

Screenshot 2023-06-20 at 11 13 24


robhyrk (approved):
Thanks, looks good!

PR #1405, opened 2023-06-20
ED checks

### LINKED ISSUE #1333: ED validation for DOT deposits and withdrawals
We should warn users if their DOT balance on Polkadot is going to get reaped

Robiquet:
closes #1333 

image
image

### DIFF
--- a/components/portfolio/WithdrawButton.tsx
+++ b/components/portfolio/WithdrawButton.tsx
@@ -23,11 +23,15 @@ const WithdrawButton = ({
   tokenSymbol,
   balance,
   foreignAssetId,
+  destinationExistentialDeposit,
+  destinationTokenBalance,
 }: {
   toChain: ChainName;
   tokenSymbol: string;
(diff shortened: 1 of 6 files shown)


components/portfolio/WithdrawButton.tsx
```diff
@@ -69,18 +75,21 @@ const createWithdrawExtrinsic = (
     "100000000000",
```
yornaath, line 75: is this ZTG? Will it vary from asset to asset?

yornaath (approved):
LGTM, just not sure if the toke transfer last param should be hardcoded?

components/portfolio/WithdrawButton.tsx
Robiquet: This is just to estimate fees. I don't think the amount transferred really has a effect, hardcoding it saves having to keep making an rpc call whenever the amount is changed

PR #1406, opened 2023-06-20
Navigation restyle

### LINKED ISSUE #1354: New navigation menu

yornaath:
Fixes #1354


yornaath:
@Robiquet I made a quick fix for the market filters not updating with url on the same page, but think the whole market filtering should be simplified in another seperate pr.

TvrtkoM:
Why is account dropdown shown covering all content on screens medium and smaller?
It can be shown like it's on desktop on all screen sizes since its with is 256px.

TvrtkoM:
there is no right margin on account button on mobile
image

TvrtkoM (approved):

PR #1412, opened 2023-06-21
Token transfer dialog

### LINKED ISSUE #1327: Token transfers
Users should be able to transfer DOT/ZTG within the app to other accounts

TvrtkoM:
close #1327


Robiquet (commented):
Looks good, just a few small things:
Bit of a Z-index issue here:
image

Asset images are pixelated, could you try `{quality={100}}`?

If it's not too much effort, it would be cool to make this clickable to set the max balance to send 
image

Robiquet:
On mobile:
Alignment issue
image

The asset doesn't seem to be clickable
image

TvrtkoM:
> Looks good, just a few small things: Bit of a Z-index issue here: image
> 
> Asset images are pixelated, could you try `{quality={100}}`?
> 
> If it's not too much effort, it would be cool to make this clickable to set the max balance to send image

I've tried to fix the z-index issue. It comes from the avatara library and I think it's not needed (could be done differently without setting z-index).

TvrtkoM:
@Robiquet I opened issue for `z-index` here. the rest is done

components/ui/AddressInput.tsx
```diff
@@ -0,0 +1,199 @@
...
+};
+
+const AddressInput: React.FC<AddressSelectProps> = ({
+  onChange,
+  error,
+  value = null,
+}) => {
+  const wallet = useWallet();
+
+  const options = useMemo<AddressOption[]>(() => {
```
yornaath, line 153: I would like to maybe reuse this in the proxy selection but with a different set of addresses, so could you refactor the AddressInput to take a prop `options: AddressOption[]` so it can be used with any set of addresses not just the wallet addresses?

yornaath (changes requested):

Robiquet (approved):
Thanks!

components/ui/AddressInput.tsx
TvrtkoM: we talked about this yesterday and we agreed that I'll integrate this component in the settings modal after you do the proxy stuff you made with Account select we use for the wallet.

yornaath (approved):

Issue #452, opened 2022-11-23
Strict Null Checks


Robiquet:
Will force us to handle null and undefined
See [docs](https://www.typescriptlang.org/tsconfig#strictNullChecks)

lsaether:
Don't use non-null assertions

PR #1456, opened 2023-06-28
#452 strict null 1

yornaath:
Implemented some more strict null checking.

**Fixes some critical client side errors on the portfolio page.**


components/trade-form/index.tsx
```diff
@@ -509,7 +524,9 @@ const TradeForm = () => {
                   validate: (value) => Number(value) > 0,
                 })}
                 onFocus={() => {
-                  setLastEditedAssetId(tradeItemState?.assetId);
+                  if (tradeItemState?.baseAssetId) {
+                    setLastEditedAssetId(tradeItemState?.baseAssetId);
```
Robiquet, line 528: This used to be `tradeItemState?.assetId`

components/trade-form/index.tsx
```diff
@@ -542,7 +561,11 @@ const TradeForm = () => {
               max="100"
               value={percentageDisplay}
               onValueChange={setPercentageDisplay}
-              onFocus={() => setLastEditedAssetId(tradeItemState?.assetId)}
+              onFocus={() => {
+                if (tradeItemState?.baseAssetId) {
+                  setLastEditedAssetId(tradeItemState?.baseAssetId);
```
Robiquet, line 566: This used to be `tradeItemState?.assetId`

lib/hooks/queries/usePortfolioPositions.ts
```diff
@@ -165,9 +168,10 @@ export const usePortfolioPositions = (
   const now = useChainTime();
 
   const { data: ztgPrice } = useZtgPrice();
-  const block24HoursAgo = Math.floor(now?.block - 7200);
+  const block24HoursAgo = Math.floor(now?.block ?? 0 - 7200);
```
Robiquet, line 171: If `now.block` is `undefined` or `null` this would return `NaN`, now it will return a negative number. I noticed there are quite a few `isNaN` checks, do any of these need to be modified?

lib/hooks/queries/usePortfolioPositions.ts
```diff
@@ -392,6 +404,7 @@ export const usePortfolioPositions = (
             );
 
             if (!price || !price24HoursAgo) {
+              console.log(poolAssetBalances24HoursAgo.query[0]?.data);
```
Robiquet, line 407: Console.log

lib/hooks/queries/usePortfolioPositions.ts
```diff
@@ -34,14 +36,15 @@ import { useTotalIssuanceForPools } from "lib/hooks/queries/useTotalIssuanceForP
...
+import { useEffect, useMemo } from "react";
 import { MarketBond, useAccountBonds } from "./useAccountBonds";
 import { useChainTime } from "lib/state/chaintime";
-import { useTradeHistory } from "./useTradeHistory";
+import { TradeHistoryItem, useTradeHistory } from "./useTradeHistory";
 import {
   ForeignAssetPrices,
   useAllForeignAssetUsdPrices,
 } from "./useAssetUsdPrice";
+import objectHash from "object-hash";
```
Robiquet, line 47: Think this can be removed

lib/hooks/queries/usePortfolioPositions.ts
```diff
@@ -692,10 +745,10 @@ export const usePortfolioPositions = (
   ]);
 
   return {
-    all: positions,
-    markets: marketPositions,
-    subsidy: subsidyPositions,
-    breakdown,
+    all: positions ?? undefined,
+    markets: marketPositions ?? undefined,
+    subsidy: subsidyPositions ?? undefined,
```
Robiquet, line 750: Seems a bit strange to do this when around line 618 the code does `?? null` why not just do `?? undefined` in the first place?

lib/hooks/queries/useTradeHistory.ts
```diff
@@ -95,7 +96,19 @@ type MarketHeader = {
   categories: { name: string }[];
 };
 
-export const useTradeHistory = (address: string) => {
+export type TradeHistoryItem = {
+  marketId: MarketId;
+  question: string;
+  assetIn: string | undefined;
```
Robiquet, line 102: Think this is the same and shorter
```suggestion
  assetIn?: string;
```

lib/hooks/queries/usePortfolioPositions.ts
```diff
@@ -458,8 +473,8 @@ export const usePortfolioPositions = (
           return acc;
         }, 0);
 
-      const calculateFifoPnl = (transactions) => {
-        let buys = [];
+      const calculateFifoPnl = (transactions: TradeHistoryItem[]) => {
```
Robiquet, line 476: Thanks for adding types :)

lib/hooks/queries/usePortfolioPositions.ts
yornaath: I will keep this resolving to NaN if current block isnt fetched. The queries that need `block24HoursAgo` are only enabled when the blocknumber is true, so in case of NaN it will wait.

Robiquet (approved):

PR #1465, opened 2023-06-29
#1419 base asset pool settings

### LINKED ISSUE #1419: Base asset of DOT market is displayed as ZTG
Screenshot 2023-06-22 at 10 31 31

yornaath:
Made some changes that I think are improvements to the pool deployment process after creating the market. The base asset bug is also fixed, and the swap fee input is the same as on the market creation page. + better loading / waiting for indexer change when deploying.

Fixes #1419


TvrtkoM (changes requested):
besides some unused code looks fine

TvrtkoM (approved):

PR #1466, opened 2023-06-29
Delay cache invalidation to wait for indexer

### LINKED ISSUE #1374: Portfolio page should show predictions right away
After I place a trade and and navigate to the portfolio page, I don't see that I have the new prediction. After a manual refresh, it shows. The portfolio page should show positions as soon as they have been confirmed on chain.

Robiquet:
closes #1374

### DIFF
--- a/lib/util/await-indexer.ts
+++ b/lib/util/await-indexer.ts
@@ -1,0 +1,9 @@
+const INDEXER_DELAY_MS = 40 * 1000;
+
+export const awaitIndexer = async (callback: () => void) => {
+  setTimeout(() => {
+    console.log("called");
+
+    callback();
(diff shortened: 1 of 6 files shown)


lib/util/await-indexer.ts
```diff
@@ -0,0 +1,9 @@
+const INDEXER_DELAY_MS = 40 * 1000;
+
+export const awaitIndexer = async (callback: () => void) => {
+  setTimeout(() => {
+    console.log("called");
```
robhyrk, line 5: can probably remove this `console.log` ?

robhyrk (approved):
looks good, just found a `console.log` that might need removing

PR #1499, opened 2023-07-05
Settings modal

TvrtkoM:


Robiquet (commented):
Could you disable fee checks for this button? There should be a property for it
image

Robiquet (approved):

PR #1501, opened 2023-07-05
Show Report/Resolution date

### LINKED ISSUE #1298: Display grace period, dispute and oracle duration on market page
After a market closes the user's funds may be locked for over a year, we should make this clearer on the market page so that they can make an informed decision. 

Perhaps we could use the "See History" modal for this

### LINKED ISSUE #1440: Show grace period for markets
It's impossible to see what the grace period of a market is. This is frustrating for advisors who have to resort to polkadot-js if they want to know if the grace period is correct. I suggest the following:

Screenshot 2023-06-26 at 10 30 31

I think maybe the user shouldn't be bothered too much with the specifics of the grace period. Instead using "Reported: (Market close) + (Grace period)" seems easier to understand. If "Reported" is too unclear, use "Open to Reports". Or something like that.

**Edit.** Replace "Resolves" with "Reported".

Robiquet:
closes #1298 closes #1440


components/markets/MarketHeader.tsx
```diff
@@ -367,6 +378,20 @@ const MarketHeader: FC<{
             dateStyle: "medium",
           }).format(ends)}
         </HeaderStat>
+        {(market.status === "Active" || market.status === "Closed") && (
```
robhyrk, line 381: for consistency with the other stats; you could do something like this for the resolution date passes: `<HeaderStat label={hasDatePassed(resolutionDateEstimate) ? "Resolved" : "Resolves"}>`

robhyrk (approved):
looks good, just left a comment with a suggested change

components/markets/MarketHeader.tsx
Robiquet: Once the market resolves the same information is shown here so I don't think we need to display it
image

PR #1512, opened 2023-07-06
market plural duration

### LINKED ISSUE #1444: Plural Duration incorrect
Should be "Month" not "Months"
image

TvrtkoM:
close #1444


robhyrk (approved):

Robiquet (commented):
Didn't know moment had this option, nice fix. Looks like the same code is duplicated in `formatDuration` which is used on the create market page, could you adjust that too?

Robiquet (approved):

PR #1524, opened 2023-07-10
Rework banner to be close to new designs

### LINKED ISSUE #1448: Header shouldn't cover banner
image
The banner should sit lower down the page or header should fade in on scroll as it did previously

yornaath:
Fixes #1448


Robiquet:
New pr for this coming

PR #1537, opened 2023-07-11
Market creation clock++ fixes

### LINKED ISSUE #1435: Market creation clock sometimes sets seconds to non-zero value
It's impossible to change the seconds, but sometimes they're suddently set to a non-zero value, which is annoying.

Screenshot 2023-06-23 at 20 28 00

### LINKED ISSUE #1436: Market creation clock default value should be midnight
It seems like the new default value of the close market clock is the current time. I think that's not very helpful. We end markets at EOD quite frequently, so it seems like 00:00 is a reasonable value.

### LINKED ISSUE #1437: Market creation clock sometimes selects wrong date
This is happening sporadically. I don't know how to reproduce this behavior. Might be related to #1435 (happened in the same session).

Screenshot 2023-06-23 at 20 29 27

yornaath:
Several improvements to market date inputs.

Fixes #1435 
Fixes #1436 
Fixes #1437 

**+ Also improves the asset ticker generation for more readability.**
**+ Also fixes some strict null checks.**

### DIFF
--- a/components/account/AccountButton.tsx
+++ b/components/account/AccountButton.tsx
@@ -88,11 +88,11 @@ const AccountButton: FC<{
     selectWallet,
     disconnectWallet,
     isNovaWallet,
-    proxyFor,
+    getProxyFor,
     realAddress,
   } = useWallet();
 
-  const proxy = proxyFor?.[activeAccount?.address];
+  const proxy = getProxyFor(activeAccount?.address);
 
   const accountModals = useAccountModals();
   const { locationAllowed, isUsingVPN } = useUserLocation();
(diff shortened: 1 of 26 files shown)


Robiquet:
Can you pull in the create page fix? I'm getting this 
image

Robiquet:
> Can you pull in the create page fix? I'm getting this
> image

I think this is a new bug, the page crashes when there is no wallet connected

TvrtkoM:
was just trying to create a market by using proxy account and it seems the create market extrinsic parameters are not being used correctly

image

the real address should be address of the real account for proxy
image

TvrtkoM:
@yornaath can you please fix the strict null check errors in `components/editor/input/asnwers/` folder?

TvrtkoM (changes requested):
@yornaath can you please fix the strict null check errors in components/editor/input/asnwers/ folder?

TvrtkoM:
trying to create a market without a proxy setting uses `proxy` extrinsic
image

TvrtkoM (approved):

PR #1563, opened 2023-07-14
Input component

### LINKED ISSUE #1481: Scrolling changes values in "Deploy Market" view
If you hover over the price of an asset and then scroll up or down, the value is incremented/decremented. We don't want this. Note that this makes it possible to set the price to negative values.

Screenshot 2023-07-03 at 14 01 30

TvrtkoM:
close #1481


components/ui/Input.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const inputRef = React.useRef<HTMLInputElement>(null);
+  const [focused, setFocused] = React.useState(false);
+
+  let { className, ...restProps } = props;
+  className = `${defaultClassName} ${className ?? ""}`;
+  const isNumber = props.type === "number";
+
+  if (isNumber) {
+    className += " font-mono";
+  }
```
yornaath, line 21: I would maybe rather do this since its a more standard approach in our code base than mutating the classname strings directly.

```ts
  let { className, ...restProps } = props;
  const isNumber = props.type === "number";
  className = `${defaultClassName} ${className ?? ""} ${isNumber ? "font-mono" : ""}`;
```

But I dont have a big problem with it tbh. Just a suggestion.

yornaath (approved):

components/ui/Input.tsx
```diff
@@ -0,0 +1,90 @@
...
+  (props, ref) => {
+    const inputRef = React.useRef<HTMLInputElement | null>(null);
+    const [focused, setFocused] = React.useState(false);
+
+    let { className, ...restProps } = props;
+    className = `${defaultClassName} ${className ?? ""}`;
+    const isNumber = props.type === "number";
+
+    if (isNumber) {
+      className += " font-mono";
```
Robiquet, line 18: Don't think we should be using the mono font anymore, @mk26071980 should be able to confirm

Robiquet (approved):

Issue #1451, opened 2023-06-27
SSG Markets cards


Robiquet:
The market cards are marking a number of calls (get identity, participants, metadata) client side that we could move to server side for faster loading

PR #1569, opened 2023-07-18
SSG market cards

Robiquet:
#1451


lib/gql/display-name.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const getDisplayName = async (
+  sdk: Sdk<FullContext>,
+  addresses: string[],
+) => {
+  const identities = await Promise.all(
+    addresses.map((address) => sdk.api.query.identity.identityOf(address)),
+  );
+
+  const textDecoder = new TextDecoder();
```
yornaath, line 11:
```suggestion
 const names: (string | null)[] = identities.map(
    (identity) => identity.unwrapOr(null)?.info?.display?.toString() ?? null,
  );
```

lib/gql/display-name.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  const names: (string | null)[] = identities.map((identity) =>
+    identity.isNone === false
+      ? textDecoder.decode(
+          (identity.value.get("info") as any).get("display").value,
+        )
+      : null,
+  );
```
yornaath, line 19: I think maybe this would be a bit more performant and more idiomatic to polkadotjs?
```suggestion
  const identities = await sdk.api.query.identity.identityOf.multi(addresses);
  const names: (string | null)[] = identities.map(
    (identity) => identity.unwrapOr(null)?.info?.display?.toString() ?? null,
  );
```

yornaath (changes requested):
Small change requests but LGTM other than that

lib/gql/display-name.ts
Robiquet: I think I did something like this before but it doesn't handle all characters correctly like Tvrtko's full name

yornaath (approved):

PR #1591, opened 2023-07-21
talisman wallet classes

### LINKED ISSUE #1460: Use wallets classes from a library
Code under `lib/wallets` should be removed in favor of using same kind of classes from a library e.g. talisman

TvrtkoM:
close #1460


lib/state/wallet.ts
```diff
@@ -207,12 +215,36 @@ let accountsSubscriptionUnsub: VoidFunction | undefined | null;
...
+  const enablePoll = async (): Promise<void> => {
+    await cryptoWaitReady();
+    try {
+      const extension = await poll(
+        async () => {
+          await wallet.enable(DAPP_NAME);
+          return wallet;
+        },
+        {
+          intervall: 66,
```
Robiquet, line 236: Should  `intervall` be `interval`?

TvrtkoM: it's `intervall` in the api

Issue #266, opened 2022-09-14
Market close should have the time and timezone


Nathanpmyoung:
The close time should be local times but need to work correctly.

maltekliemann:
I'm bumping this. It would be really good to have the following features with regards to dates:

- Let the user specify the timezone, but always use UTC by default.
- Specify all dates with a timezone using a format like ISO8601 or some watered-down version of that
- (I'm dreaming here) Allow market creators to specify dates in the description in a way that the date changes according to the user's choice of timezone (so if I write: "The market will resolve to YES if Donald Trump tweets before 2022-04-01T00:00Z" and the user's timezone is CEST, then they see "The market will resolve to YES if Donald Trump tweets before 2022-04-01T01:00CEST").

maltekliemann:
And while we're at it, can we use a 24h clock?

Robiquet:
We should be storing dates in UTC and displaying them in the timezone and format based on the user's browser locale using APIs such as [Intl](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat). If that's not happening currently let us know and we can fix that. 

I don't think displaying dates in UTC timezone and ISO formatting isn't a great idea as it forces users to look up their timezone in relation to UTC/ do maths. ISO formatting can also be quite different for what certain users are used to

That being said, I can see a use case for specifying timezones on market creation as it could save some manual conversion when creating markets outside of your country.

maltekliemann:
Bumping this because it's a really annoying problem.

lsaether:
@TvrtkoM Can you complete this by the end of the week?

TvrtkoM:
addded ability for user to set the timezone here https://github.com/zeitgeistpm/ui/pull/1621

@maltekliemann @Robiquet can you please see is this going to be fine?

date is set in specified timezone when market is created, but it's stored in UTC. Other date displays e.g. on market page are shown in UTC

![Image](https://github.com/zeitgeistpm/ui/assets/551965/984f4bc3-f479-43fd-8f5e-bac0a1983a9c)

In my opinion this should be a global setting (that's set in user settings modal) and all dates displayed in the app should use this locale.

maltekliemann:
For now, this is fine. However...

> In my opinion this should be a global setting (that's set in user settings modal) and all dates displayed in the app should use this locale.

[...]

PR #1621, opened 2023-07-27
allow market creators to specify timezone

TvrtkoM:
#266


Robiquet:
Is it possible to default to local timezone with something like `new Date().getTimezoneOffset()`? @mk26071980 could you provide a design for this?

TvrtkoM:
> Is it possible to default to local timezone with something like `new Date().getTimezoneOffset()`? @mk26071980 could you provide a design for this?

it is possible with `moment.tz.guess()` but the issue mentiones the default should be UTC

Robiquet (commented):
Seems like something strange has happened to the drop downs.

Looked like this before:
image

Now it looks like:
image

Robiquet (approved):
Actually ignore the above, looks like just a safari thing

PR #1672, opened 2023-08-08
Create page loader

### LINKED ISSUE #1385: Handle create page loading
Create page can hang in this state for some time even when there is no data to load. We should try to reduce this and handle it better when loading time is needed 
image

### LINKED ISSUE #1666: Show loader on create market page

Robiquet:
closes #1385 
closes #1666

### DIFF
--- a/components/create/editor/Editor.tsx
+++ b/components/create/editor/Editor.tsx
@@ -62,8 +62,6 @@ export const MarketEditor = () => {
     input,
     fieldsState,
     mergeFormData,
-    isTouched,
-    reset,
   } = editor;
 
   const chainTime = useChainTime();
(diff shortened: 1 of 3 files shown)


components/create/editor/Editor.tsx
```diff
@@ -101,528 +100,545 @@ export const MarketEditor = () => {
...
-      </div>
-
-      <div className="mb-8 md:mb-12">
-        {isWizard && (
-          <WizardStepper
-            steps={steps}
-            current={currentStep}
-            onChange={(step) => setStep(step)}
+    <>
+      {isLoaded === false && (
```
Robiquet, line 107: Diff is pretty horrible, I've just wrapped everything in `<></>` and added a loader here

TvrtkoM (approved):

pages/create.tsx
```diff
@@ -1,9 +1,5 @@
+import MarketEditor from "components/create/editor/Editor";
 import { NextPage } from "next";
-import dynamic from "next/dynamic";
-
-const MarketEditor = dynamic(() => import("components/create/editor/Editor"), {
```
Robiquet, line 4: @yornaath do you remember this being needed for any reason? Removing it makes loading quite a bit faster

PR #1673, opened 2023-08-08
decrease spacing on market filters

### LINKED ISSUE #1660: Reduce padding and empty space on the filters subheader

### LINKED ISSUE #1665: Header misalignment on create page
image

TvrtkoM:
close #1660 
close #1665


lsaether:
Screenshot_20230810_074403

What happened to the margins?

lsaether:
Screenshot_20230810_074606

What happened to the margins?

yornaath:
The filter seems just a tad constricted, I like that its "tight" to the top bar, but the padding feels a bit off compared to the rest of the site.

TvrtkoM:
@lsaether I made spacing and width a bit better now, but think final styling should be done as a separate task once responsive design is done so we can define new responsive breakpoints for the whole app that way.

@yornaath I increased vertical margin between filter rows, not sure is that fine now?

yornaath:
Personally I would expand the margin even a bit more, but cant find the new designs so might be in line with them? 
Screenshot 2023-08-21 at 09 52 25

TvrtkoM:
> Personally I would expand the margin even a bit more, but cant find the new designs so might be in line with them? Screenshot 2023-08-21 at 09 52 25

There is no design for this. The related issue was to decrease this spacing between filter rows. I'm not sure now what to do, increase the margin or decrease.

Robiquet (approved):

PR #1737, opened 2023-08-21
Front page redesign

### LINKED ISSUE #1607: Main page new layout for banner and News section
Replace the carousel with the main banner and news section that links out to blog posts.

yornaath:
Rework the banner to the new designs.
Fixes #1607 

@lsaether Need link for the Learn More button.


Robiquet (commented):
Look good! Might be worth generating a blur image for this as it took quite a while to load. Also, tests are failing, could you check if they could be modified to make them work?
image

components/front-page/HeroBanner.tsx
```diff
@@ -0,0 +1,98 @@
...
+            </div>
+          </div>
+        </div>
+        <div className="w-full h-64 md:h-auto relative rounded-lg overflow-hidden mb-8 md:mb-0">
+          <Image
+            alt="Futuristic City Image"
+            fill={true}
+            sizes="100vw"
+            className="object-cover"
+            src="https://cdn.discordapp.com/attachments/826371897084215376/1138829878188327043/image.png"
```
Robiquet, line 92: Why reference via discord? Not sure if image optimisation is working

TvrtkoM:
on smaller screens there is white space on the right side and horizontal scrollbar appears

mk26071980:
Pls, make the space between the top navbar and banner section 35 pixels.
Card row should contain 4 cards for the screens wider than 1728 px

components/markets/MarketScroll.tsx
```diff
@@ -32,39 +33,27 @@ const MarketScroll = ({
...
-      const newScroll = prev - moveSize;
-
-      return newScroll < scrollMin ? scrollMin : newScroll;
-    });
+    setPageIndex(pageIndex - 1);
   };
 
-  const hasReachedEnd = scrollMax - containerWidth - scrollLeft === 0;
-  const leftDisabled = scrollLeft === 0;
+  const pageDelta = markets.length / cardsShown;
```
Robiquet, line 52: What's the reason for rewriting this? I'm noticing some strange behaviour
image

pages/index.tsx
```diff
@@ -38,94 +47,115 @@ export async function getStaticProps() {
...
       featuredMarkets: featuredMarkets ?? [],
       trendingMarkets: trendingMarkets ?? [],
+      bannerPlaceholder: bannerPlaceholder.base64 ?? "",
       categoryCounts: categoryCounts,
       categoryPlaceholders: categoryPlaceholders.map((c) => c.base64) ?? [],
-      bannerPlaceHolders: bannerPlaceHolders.map((c) => c.base64) ?? [],
+      newsImagePlaceholders: newsImagePlaceholders.map((c) => c.base64) ?? [],
       stats,
+      ztgHistory,
+      chainProperties: chainProperties.toHuman(),
```
Robiquet, line 94: `toHuman` can do weird things like adding commas to big numbers. Might just be better to pull out what you need, it will be less data send to the client too

Robiquet (commented):
Looks good, just a few other things:
Header no longer sticks to the top
image
There is also a bit of gap when the menu is open and the content behind it is scrollable
image

pages/index.tsx
Robiquet: I find `toPrimitive` is often a better option if you wanted to go down that route

components/markets/MarketScroll.tsx
yornaath: Yeah I had to rewrite a bit because when we use overflow scroll the cards get clipped when hovering and its scale is increased outside the boundaries of the container. So I am using translation to offset and opacity to show them. Its a bit more tricky, but a nice thing about it is that the cards fade in from the edges on bigger screens.

yornaath: Should be fixed

pages/index.tsx
yornaath: Should also be fixed

components/menu/index.tsx
```diff
@@ -24,7 +24,7 @@ const AccountButton = dynamic(() => import("../account/AccountButton"), {
 const TopBar = () => {
   return (
     <div
-      className={`w-full py-3.5 sticky top-0 z-40 transition-all duration-300 bg-black h-topbar-height`}
+      className={`w-full py-3.5 fixed top-0 z-40 transition-all duration-300 bg-black h-topbar-height`}
```
robhyrk, line 27: This seems to be unrelated to the css change but I'm posting here since it's relevant: previously the scroll was disabled on mobile so that when the hamburger menu was open the user couldn't scroll up/down. Perhaps this could be opened as a separate issue.

components/front-page/GettingStartedSection.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+const GettingStartedSection = () => {
+  return (
+    <>
+      <div className="w-full" data-testid="learnSection">
+        <h2 className="sm:col-span-3 text-center sm:text-start mb-6">
+          Getting Started
+        </h2>
+        <div className="flex flex-col md:flex-row gap-4">
+          <div className="flex-1">
```
robhyrk, line 13: Some alignment issues on the 905px breakpoint

Screenshot 2023-09-05 at 9 21 35 PM

components/front-page/BgBallFx.tsx
```diff
@@ -0,0 +1,84 @@
...
+      damping: 30,
+      restDelta: 0.001,
+    }),
+    [-1400, 1400],
+    [-50, 50],
+  );
+
+  const oy = useTransform(() => osy.get() + omy.get());
+
+  return (
```
robhyrk, line 62: Not sure where this is happening exactly but there is an error in the console happening at `WatchHow.tsx`: `Encountered two children with the same key, ` `. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.`

components/front-page/News.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export const NewsSection = ({
+  news,
+  imagePlaceholders,
+}: {
+  news: News[];
+  imagePlaceholders: string[];
+}) => {
+  return (
+    <div className="mb-12">
```
robhyrk, line 14: Hover effect slightly cuts off card near the mid-range breakpoint (below 905px)

Screenshot 2023-09-05 at 9 40 50 PM

components/front-page/PopularCategories.tsx
```diff
@@ -28,10 +28,10 @@ const Category = ({
 }) => {
   return (
```
robhyrk, line 29: I can't seem to be able to scroll through horizontally the popular categories anymore. (tested on desktop only)

Screenshot 2023-09-05 at 9 47 24 PM

lib/hooks/events/useHasMounted.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useState } from "react";
```
robhyrk, line 1: just checking if this hook is being used anywhere as I couldn't seem to find it

robhyrk (changes requested):
Looking good!

components/markets/market-card/index.tsx
```diff
@@ -241,14 +225,29 @@ const MarketCard = ({
...
         <Link
           href={`/markets/${marketId}`}
-          className="flex flex-col flex-1 gap-4"
+          onClick={(e) => {
+            if (disableLink) {
+              e.preventDefault();
+              return;
+            }
+          }}
+          className={`flex flex-col flex-1 gap-4 ${
```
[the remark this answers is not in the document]
yornaath: Yeah that error cant easily be removed since the client will actually render a bit different from the server, it was related to the issue I was talking about earlier. See this bit of code: https://github.com/zeitgeistpm/ui/pull/1737/files#diff-8a23978fb2a961548d1b3b3c2a25caa7d32c079a3411b806ee395a86288d007aR109-R132

components/front-page/GettingStartedSection.tsx
yornaath: Should be better in the newest build.

components/front-page/News.tsx
yornaath: Yeah was the same for market cards. I just disabled the hover effects for mobile since its not really necessary when its touch event based, actually kinda looks weird when it gets enlarged when clicked/touched as opposed to hovered.

components/front-page/PopularCategories.tsx
yornaath: Touch events(dragging) seems to work tough so dont thin this will be an issue on mobile.

components/menu/index.tsx
robhyrk: i think we can leave it as is and open an issue for it if deemed necessary.

components/front-page/News.tsx
robhyrk: Good call!

components/front-page/PopularCategories.tsx
robhyrk: Okay, sounds good

robhyrk (approved):

PR #1741, opened 2023-08-22
Price fetching cache

### LINKED ISSUE #1697: Foreign asset price fetching
Sometimes price fetching is failing and making the page crash. We should make the code more robust and investigate better ways to fetch prices. Maybe can use edge functions to cache

Robiquet:
closes #1697


pages/api/usd-price-edge.ts
```diff
@@ -0,0 +1,33 @@
+import { NextResponse } from "next/server";
```
yornaath, line 1: Maybe remove this file if edge didnt work for this?

PR #1746, opened 2023-08-24
Market page redesign

### LINKED ISSUE #1610: New market page

yornaath:
Fixes #1610


components/markets/SimilarMarketsSection.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const hasSimilarMarkets = Boolean(similarMarkets?.data?.length);
+
+  const stats = useMarketsStats(
+    similarMarkets?.data?.map((m) => m.marketId) ?? [],
+    {
+      enabled: hasSimilarMarkets,
+    },
+  );
+
+  const isLoading =
```
Robiquet, line 27: I think this can be simplified quite a bit. Think if `isFetched` is true you can assume it's not loading anymore. https://tanstack.com/query/v4/docs/react/reference/useQuery

components/outcomes/ScalarDisputeBox.tsx
```diff
@@ -83,17 +87,31 @@ const ScalarDisputeBox = ({
...
+      onBroadcast: () => {},
       onSuccess: () => {
-        notificationStore.pushNotification("Outcome Disputed", {
-          type: "Success",
-        });
-        onSuccess?.();
+        if (onSuccess) {
+          onSuccess?.({
+            type: market.scalarType as ScalarRangeType,
+            scalar: new Decimal(scalarReportValue).mul(ZTG).toFixed(0),
```
Robiquet, line 95: Is it correct to multiply here? I thought it would be the other way around

components/outcomes/ScalarReportBox.tsx
```diff
@@ -54,11 +60,18 @@ const ScalarReportBox = ({
...
+      onBroadcast: () => {},
       onSuccess: () => {
-        notificationStore.pushNotification("Outcome Reported", {
-          type: "Success",
-        });
-        onReport?.();
+        if (onReport) {
+          onReport?.({
+            type: market.scalarType as ScalarRangeType,
+            scalar: new Decimal(scalarReportValue).mul(ZTG).toFixed(0),
```
Robiquet, line 68: Same here, not sure if multiplication is correct

components/outcomes/ScalarReportBox.tsx
```diff
@@ -72,47 +85,107 @@ const ScalarReportBox = ({
...
+                minWidth: expandedInfoToggled ? digits * 18 : digits * 12,
+              }}
+            >
+              <div className="self-start flex sm:hidden md:flex flex-1 lg:hidden">
+                Scalar range:
+              </div>
+              <div className="whitespace-nowrap">
+                {expandedInfoToggled
+                  ? bounds[0].toNumber()
+                  : new Intl.NumberFormat("en-IN", {
```
Robiquet, line 160: These should be "en-US"

components/ui/Table.tsx
```diff
@@ -415,7 +415,9 @@ const Table = ({
                     .map((column, index) => (
                       <th
                         key={index}
-                        className={`${getHeaderClass(column)} ${
+                        className={`${getHeaderClass(
+                          column,
+                        )} border-b-2 border-purple-100 ${
```
Robiquet, line 420: Apparently the purple border was a design mistake

lib/hooks/queries/useMarketsStats.ts
```diff
@@ -7,6 +7,9 @@ export const marketsStatsRootQuery = "marketsStats";
 
 export const useMarketsStats = (
   marketIds: number[],
+  opts?: {
```
Robiquet, line 10: Is this needed? We are already checking that `marketIds.length > 0`

tailwind.config.js
```diff
@@ -181,6 +192,7 @@ module.exports = {
         success: "#E8FFE4",
         info: "#E4F5FF",
         error: "#FFE6E4",
+        "ztg-pink": "#DC056C",
```
Robiquet, line 195: Can't see this used anywhere

styles/index.css
```diff
@@ -214,6 +214,25 @@ h6 {
   @apply px-9 mx-auto max-w-screen-3xl;
 }
 
+.invis-scrollbar::-webkit-scrollbar {
```
Robiquet, line 217: Is this used any where?

Robiquet (commented):
Thanks, looking good!

components/markets/MarketContextActionOutcomeSelector.tsx
```diff
@@ -0,0 +1,133 @@
...
+      const name = market.categories?.[getIndexOf(option)].name ?? "";
+      return { option, name };
+    });
+
+    options = fuzzysort
+      .go(search, opts, { key: "name" })
+      .map((result) => result.obj.option);
+  }
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
yornaath: I think this is better because the state regarding if its open or not is set in multiple places, so would have to be called every time the state explicetly changes to !open. Also the `MarketContextActionOutcomeSelector` is rendered containing its search state even when its closed, so clearing search on unmount would also not work as expected.

tailwind.config.js
yornaath: Ah yeah I think I added this on the wrong pr, it was meant to be on the new front page design pr

styles/index.css
yornaath: No can be removed, dont need it after all

components/outcomes/ScalarDisputeBox.tsx
yornaath: Yeah I think you are right since the `scalarReportValue` is the raw value from the input I think

Robiquet (commented):
I think this comment still needs to be addressed
https://github.com/zeitgeistpm/ui/pull/1746/files#r1309949300

Robiquet (approved):

PR #1775, opened 2023-09-07
Update issue templates

yornaath:


robhyrk:
@yornaath @Robiquet should we discuss or close this out as well?

PR #1782, opened 2023-09-11
Map error type to message

### LINKED ISSUE #1763: Incorrect error message when connecting wallet
image

Robiquet:
closes #1763

### DIFF
--- a/components/account/WalletSelect.tsx
+++ b/components/account/WalletSelect.tsx
@@ -63,9 +63,10 @@ const WalletSelect = () => {
               )}
               {hasError && (
                 <div className="text-vermilion ml-auto  text-ztg-12-120 w-ztg-275">
-                  {error.type === "NoAccounts"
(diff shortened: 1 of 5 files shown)


components/account/WalletSelect.tsx
```diff
@@ -63,9 +63,10 @@ const WalletSelect = () => {
               )}
               {hasError && (
                 <div className="text-vermilion ml-auto  text-ztg-12-120 w-ztg-275">
-                  {error.type === "NoAccounts"
```
yornaath, line 66: What does this actually change? Since error type should only be one of two types it seems like the old code should cover all cases?

Robiquet: I suspect the types are probably wrong and not all errors are accounted for. Even so, I don't think the previous code is good practice as it assumes that there will only ever be 2 error messages

yornaath (approved):

PR #1786, opened 2023-09-11
Fix: adjust font optimization

### LINKED ISSUE #1785: Troubleshoot typography loading

robhyrk:
fixes #1785

### DIFF
--- a/layouts/DefaultLayout.tsx
+++ b/layouts/DefaultLayout.tsx
@@ -3,7 +3,6 @@ import { useResizeDetector } from "react-resize-detector";
 import Image from "next/image";
 import dynamic from "next/dynamic";
 
-import { ContentDimensionsProvider } from "components/context/ContentDimensionsContext";
 import TopBar from "components/menu";
 import Footer from "components/ui/Footer";
 import NotificationCenter from "components/ui/NotificationCenter";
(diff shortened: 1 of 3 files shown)


Robiquet (approved):

Robiquet (changes requested):
Sorry... just noticed that fonts aren't loading here (Get started modal). Probably because it's using some sort of portal so that the modal is outside of the element where the fonts were added
image

layouts/DefaultLayout.tsx
```diff
@@ -47,24 +46,16 @@ const DefaultLayout: FC<PropsWithChildren> = ({ children }) => {
 
   return (
     <div
-      className={`relative min-h-screen justify-evenly overflow-x-hidden ${
+      className={`${inter.variable} ${kanit.variable} ${
```
yornaath, line 49: I would probably put this on the body element in _document. Should fix the issue @Robiquet reported

yornaath (changes requested):

robhyrk:
The nextjs docs didn't explain it the best but I found a workaround because for some reason font optimization doesn't load when called in `_document` so I had to add in` _app`, but then there was the issue of not being able to access the `body` tag directly so I ended up adding in the fonts dynamically.

Robiquet:
I'm still seeing the shifting like we have on staging, I wonder if we can move the modal portal to it get the fonts?

Robiquet (approved):

yornaath (approved):

PR #1827, opened 2023-09-27
Fix sidebar not being sticky on market page

yornaath:


Robiquet (changes requested):
Looks like there are some issues with changing the global layout 
image

Robiquet (commented):
I'm still seeing the same issue I mentioned above, I think we just need to add a background to the market filters section

Robiquet (approved):

PR #1831, opened 2023-09-28
Show expected pool ownsership after trade

### LINKED ISSUE #1756: Show expected % of pool ownership on Pool Join modal

yornaath:
Fixes #1756


Robiquet (changes requested):
I think we should scale down the spacing for mobile a bit or display the new row inline like the outcome rows. Might also be good to hide the buy/sell buttons when the modal opens
image

Robiquet (approved):

PR #1846, opened 2023-10-03
Alerts, show market ready to report for oracle

### LINKED ISSUE #61: Allow people to flag markets that need to be resolved + notifications
The market creator and admin(s) would be alerted via in-app notifications.

yornaath:
Partially Fixes #61 
- [x] Reportable market alert
- [x] Redeemable market alert
- [x] Functionality to mark alerts as read(persisted client side). _For when we need to add alerts that aren't automatically removed by performing a action(like redeeming tokens or reporting the outcome)._


lib/hooks/queries/useReadyToReportMarkets.ts
```diff
@@ -0,0 +1,45 @@
...
+          where: {
+            oracle_eq: account,
+            status_eq: MarketStatus.Closed,
+          },
+        });
+
+        let readyToReportMarkets = (
+          await Promise.all(
+            closedMarketsForAccount.markets.map(async (market) => {
+              const stage = await sdk.model.markets.getStage(market);
```
Robiquet, line 26: Could we instead just use market status? That way it will just be 2 requests rather than a request for every asset

yornaath: No status of a market will be closed also when its in graceperiod, so we have to determine the more granular stage of the market, if its in oracle reporting or open reporting. The async function `model.markets.getStage` doesnt do a lot of heavy lifting, it just fetches the current on chain time and block, but I can pass that as a param so its only fetched once up front. Other than that it shouldnt make any more requests for closed markets.

Robiquet: Ah yeah, probably would be best to pull the calculation stuff out of the function so we can call it separately if we already have all of the info needed. I think it's ok for now though

robhyrk:
Could you link the alerts to it's respective area within the app? (i.e. when the user clicks the alert to redeem tokens it goes to the portfolio page?). Might be more useful for newer users who don't know where things are, and would also save users a few clicks.

I think it would be good to have some sort of "x" to dismiss the notification as well.

Robiquet (changes requested):
Looks like an unintentional styling change here
image

Robiquet (approved):

PR #1857, opened 2023-10-05
Add missing form mapping

### LINKED ISSUE #1854: Timezone selection doesn't work
It's not possible to change the timezone on the drop down
image

Robiquet:
closes #1854 
Need to test once battery station is back


robhyrk (changes requested):
Dropdown works now but there seems to be an error when trying to publish the market on Battery Station: `Extrinsic predictionMarkets.createCpmmMarketAndDeployAssets expects 11 arguments, got 10`

robhyrk (approved):

PR #1862, opened 2023-10-06
Setting changes

### LINKED ISSUE #1675: Allow users to select fee paying asset
Probably somewhere in user settings the user should be able to set the asset they want to pay network fees with

### LINKED ISSUE #1729: Add info tooltip to proxy execution setting

Robiquet:
closes #1675 
closes #1729

### DIFF
--- a/components/settings/FeePayingAssetSelect.tsx
+++ b/components/settings/FeePayingAssetSelect.tsx
@@ -62,8 +62,8 @@ const FeePayingAssetSelect = () => {
         <label className="font-bold">Select asset to pay network fees</label>
         {showSaved && (
           <div className="flex gap-2 items-center ml-auto">
-            <Check size={18} className="text-green-500" />
-            <div>Saved</div>
+            <Check size={16} className="text-green-500" />
+            <div className="text-sm">Saved</div>
           </div>
         )}
       </div>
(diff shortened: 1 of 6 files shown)


components/settings/OtherSettingsForm.tsx
robhyrk, line 1: This might be a UX issue: a user can click on `Enable Proxy Execution`, select an account that is not a proxy, uncheck `Enable Proxy Execution`, and then hit save. I suppose this doesn't actually do anything? Maybe you could remove the address input if `Enable Proxy Execution` is not enabled?

robhyrk (changes requested):

components/settings/FeePayingAssetSelect.tsx
```diff
@@ -0,0 +1,91 @@
...
+import {
+  FOREIGN_ASSET_METADATA,
+  findAssetImageForSymbol,
+} from "lib/constants/foreign-asset";
+import { useAllAssetMetadata } from "lib/hooks/queries/useAssetMetadata";
+import useFeePayingAssetSelection from "lib/state/fee-paying-asset";
+import { useMemo, useState } from "react";
+import { Check } from "react-feather";
+
+const isSupportedAsset = (id: number) => {
```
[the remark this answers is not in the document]
Robiquet: I think it should be filtered as it's not inside `FOREIGN_ASSET_METADATA`

components/settings/OtherSettingsForm.tsx
Robiquet: Yeah I'm not sure what the expected behaviour is for this, I'm wondering if we even need the checkbox. Maybe put this as a separate issue and we can discuss?

PR #1871, opened 2023-10-10
Web3auth

### LINKED ISSUE #1629: Investigate Web3Auth options
There are a few options for username+password signup that provides private keys. We should investigate these options to see if they are compatible to reduce onboarding friction

https://web3auth.io/docs/how-web3auth-works
https://web3auth.io/docs/sdk/helper-sdks/providers/common

https://magic.link/ (expensive but has polkadot support)

robhyrk:
closes #1629 

- [x] create web3 auth wallet hook
- [x] add web3auth wallet details to supported wallet instance
- [x] import and setup sign in method using web3 auth within wallet select
- [x] setup and pass through web3 auth wallet provider for signing transactions
- [x] update types
- [x] update user config store with web3 auth to persist state after refresh
- [x] explore why profile page is not loading history for web3auth wallet transactions
- [x] test (sign in, sign out, switch wallets, sign transactions, view profile page)


lib/state/util/web3auth-config.ts
```diff
@@ -0,0 +1,67 @@
...
+    alt: logoAlt,
+  },
+});
+
+const chainConfig = {
+  chainNamespace: CHAIN_NAMESPACES.OTHER,
+  chainId: "0x1",
+  rpcTarget: "https://rpc.polkadot.io/",
+  displayName: "Polkadot Mainnet",
+  blockExplorer: "https://explorer.polkascan.io/",
```
yornaath, line 40: Not entirely sure what these configs do, and doesnt seem to be needed to sign transactions. But maybe they should be configured to zeitgeist?

lib/state/wallet.ts
```diff
@@ -33,10 +39,14 @@ export type UseWallet = WalletState & {
...
   isNovaWallet: boolean;
+  /**
+   * Name of the wallet being used
+   */
+  walletId?: string;
   /**
    * Get the active signer for transactions. Is either the real account or the proxy account.
    */
-  getSigner: () => KeyringPairOrExtSigner | undefined;
+  getSigner: (walletId?: string) => KeyringPairOrExtSigner | undefined;
```
yornaath, line 49: This functions signature doesnt need the walletId param anymore

lib/util/tx.ts
```diff
@@ -110,6 +111,7 @@ export const signAndSend = async (
   signer: KeyringPairOrExtSigner,
   cb?: GenericCallback,
   foreignAssetNumber?: number,
+  walletId?: string | undefined,
```
yornaath, line 114: This param is unused

lib/state/wallet.ts
```diff
@@ -321,33 +360,64 @@ if (initialWalletId) {
...
+    web3auth.status === "not_ready" && (await web3auth.initModal());
+    await web3auth.connect();
+
+    if (web3auth.provider) {
+      const getKeypair = async (provider) => {
+        await cryptoWaitReady();
+        const privateKey = await provider.request({
+          method: "private_key",
+        });
+        const keyring = new Keyring({ ss58Format: 42, type: "sr25519" });
```
yornaath, line 379: Think the correct ss8 format is 73, same as you do in encoding of ztgAddress

yornaath (changes requested):

lib/state/wallet.ts
```diff
@@ -321,33 +360,64 @@ if (initialWalletId) {
...
+  const loadWeb3AuthWallet = async () => {
+    if (!web3auth) {
+      return;
+    }
+
+    web3auth.status === "not_ready" && (await web3auth.initModal());
+    await web3auth.connect();
+
+    if (web3auth.provider) {
+      const getKeypair = async (provider) => {
```
yornaath, line 374: Please type this param `(provider: IProvider)`
Found in `import { IProvider } from "@web3auth/base";`

yornaath (changes requested):

lib/state/util/web3auth-config.ts
robhyrk: From my understanding it's used to generate the keyring pair for users. I'm not sure if it's compatible with Zeitgeist specific settings but I can try those instead and see if it works.

lib/state/wallet.ts
Robiquet: Ideally we should pull this from chain config, the property is `ss58Prefix`

yornaath (approved):

PR #1884, opened 2023-10-12
Get tokens button

yornaath:


Robiquet (commented):
Would it be possible to render a skeleton or some placeholder for this button to avoid layout shift? 
image

Robiquet (approved):

PR #2010, opened 2023-11-24
Misc fixes

yornaath:
- [ ] Top bar alerts not updating when revealing or voting.
- [ ] Enable click to open file browser when restoring court vote/reveal salt. Courtesy of @Chralt98 
- [ ] Disable court case page static generation when court flag is false.
- [ ] Fix liquidity input bug in market creation


robhyrk (approved):
Looks good!

PR #2018, opened 2023-11-28
Implement ability to re cast votes during voting period if already voted

yornaath:


components/court/CourtVoteForm.tsx
```diff
@@ -142,8 +145,10 @@ export const CourtVoteForm: React.FC<CourtVoteFormProps> = ({
                 <code className="mb-4 block text-xs">
                   <span>
                     vote_item = VoteItem::Outcome(OutcomeReport::Categorical(
```
robhyrk, line 147: Is this supposed to be here as a string of text in this `span` tag? Looks like code from somewhere else but maybe I'm not sure of the context

robhyrk (changes requested):

robhyrk (approved):

PR #2025, opened 2023-11-29
Improve search bar UX

yornaath:
Tried to improve the search bar UX a bit.

### DIFF
--- a/components/markets/MarketSearch.tsx
+++ b/components/markets/MarketSearch.tsx
@@ -1,7 +1,9 @@
+import { Transition } from "@headlessui/react";
 import { MarketStatus } from "@zeitgeistpm/indexer";
 import { useMarketSearch } from "lib/hooks/queries/useMarketSearch";
 import Link from "next/link";
-import { useEffect, useRef, useState } from "react";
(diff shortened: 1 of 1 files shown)


components/markets/MarketSearch.tsx
```diff
@@ -30,77 +33,97 @@ const MarketSearch = () => {
...
+            setSearchTerm(event.target.value);
+          }}
+          onFocus={() => {
+            if (searchTerm?.length > 0) {
+              setShowResults(true);
+            }
+          }}
+        />
+        {showSearch && (
+          <button
```
robhyrk, line 54: I think it would be helpful for the user to see on desktop screens that search is gathering results with a loading spinner too. Perhaps the search icon could be replaced with a loading spinner while that state is occuring?

 
Screenshot 2023-11-29 at 11 54 32 AM

robhyrk (changes requested):

robhyrk (changes requested):
I just noticed the loading effect was missing on mobile.

yornaath:
> I just noticed the loading effect was missing on mobile.

Should be fixed

robhyrk (approved):

PR #2097, opened 2023-12-18
Disclaimer

### LINKED ISSUE #1822: Add disclaimer
If vpn user is detected, show disclaimer

Robiquet:
closes #1822


robhyrk (changes requested):
I think there may be an issue with how the persistent atom is being saved. I did some testing and I noticed the the atom being saved in storage holds the following value after accepting the user agreement `{__version: 0}`. It seems like the boolean for the `defaultValue` is not being saved. That being said, the disclaimer functionality seems to work since it checks for the truthy value of the`disclaimerStatus` object returned.

As an aside: I'm not sure if userLocation is working correctly either. I used as VPN via the US and it doesn't seem to be picking up correctly.

Robiquet:
> I think there may be an issue with how the persistent atom is being saved. I did some testing and I noticed the the atom being saved in storage holds the following value after accepting the user agreement `{__version: 0}`. It seems like the boolean for the `defaultValue` is not being saved. That being said, the disclaimer functionality seems to work since it checks for the truthy value of the`disclaimerStatus` object returned.
> 
> As an aside: I'm not sure if userLocation is working correctly either. I used as VPN via the US and it doesn't seem to be picking up correctly.

I think the `disclaimerStatus` is a quirk of `persistentAtom` I've changed it to an object to make it explicit. Regarding the location, what behaviour are you expecting? I added some debug logs to verify the location is getting picked up correctly

PR #2126, opened 2023-12-20
Court fixes

### LINKED ISSUE #2019: [Court] Cannot increase self-stake beyond 50%
**Bug**
Increasing self-stake tries to set the new self-stake instead of increasing it.

**Steps to reproduce**
1. Join Court as a juror using 50% of available ZTG
2. Click on "Increase Personal Stake"
3. Try to increase personal stake.

**Potential solution**
Consider `free_balance - locked_in_court` as available amount to stake. Instead of using that value as the total self stake, consider using that value plus the existing self stake as the new total self stake.

**Logs**
`system.account` output for the test account:
```
{
  nonce: 2
  consumers: 1
  providers: 1
  sufficients: 0
  data: {
    free: 499,999,552,319,146
    reserved: 0
    miscFrozen: 250,000,000,000,000
    feeFrozen: 250,000,000,000,000
  }
}
```

UI when trying to increase self stake beyond 50%:
image
Note that when the screenshot was taken, 50% were locked and 50% were free. The maximum value achievable using the slider is the free balance. Thus, increasing the self stake is impossible although balance is available for that.

### LINKED ISSUE #2148: [Court] Vote Count instead of vote weight shown
Screenshot 2024-01-06 at 16 12 46

Unfortunately I see the vote count (2 votes) for WELL, but I would have expected the vote weight:

`13 + 18 = 31`

Otherwise it would be confusing for the following case: You see there are 4 votes for outcome 1 and 3 votes for outcome 2. But in the end outcome 2 wins. What happened? The 3 votes together had more vote weight `18`, although more accounts (4 votes) were chosen who selected outcome 1 (but you had `13` vote weight for outcome 1 instead).

### LINKED ISSUE #2149: [Court] `Waiting for vote` seems to be buggy
Screenshot 2024-01-06 at 16 05 06

I would have not expected `Waiting for vote` in the appeal phase.

yornaath:
- [x] Take into account already staked balance when setting juror or delegated stake.
- [x] Show reported outcome on case pages for settled markets.
- [x] Use copy "set stake" instead of "increase stake" for more clarity.

Fixes #2148
Fixes #2149
Fixes #2019

### DIFF
--- a/components/court/JoinCourtAsJurorButton.tsx
+++ b/components/court/JoinCourtAsJurorButton.tsx
@@ -135,8 +135,6 @@ const JoinCourtAsJurorButton = ({ className }: { className?: string }) => {
               : "Become a Juror"}
           </h3>
 
-          {balance.toString()}
-
           <div className="mt-[20px] flex w-full flex-col items-center gap-8 text-ztg-18-150 font-semibold">
             <form
               onSubmit={handleSubmit(onSubmit)}
--- a/components/court/JoinCourtAsJurorButton.tsx
+++ b/components/court/JoinCourtAsJurorButton.tsx
@@ -14,7 +14,7 @@ import { useExtrinsic } from "lib/hooks/useExtrinsic";
 import { useSdkv2 } from "lib/hooks/useSdkv2";
 import { useNotifications } from "lib/state/notifications";
 import { useWallet } from "lib/state/wallet";
-import { useEffect, useState } from "react";
+import { useEffect, useMemo, useState } from "react";
 import { useForm } from "react-hook-form";
 import { IoIosInformation, IoIosWarning } from "react-icons/io";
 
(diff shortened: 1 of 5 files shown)


components/court/JoinCourtAsJurorButton.tsx
```diff
@@ -125,10 +131,12 @@ const JoinCourtAsJurorButton = ({ className }: { className?: string }) => {
         <Dialog.Panel className="w-full max-w-[462px] rounded-[10px] bg-white p-[30px]">
           <h3 className="mb-4">
             {connectedParticipant?.type === "Juror"
-              ? "Increase Personal Stake"
+              ? "Set Personal Stake"
               : "Become a Juror"}
           </h3>
 
+          {balance.toString()}
```
Robiquet, line 138: Think this should have a balance label and units, or is this just debug code?

Robiquet (approved):

PR #2152, opened 2024-01-09
Search for markets by searchterm "id:number"

yornaath:
Make it possible to search for markets by searchterm "id:number" like "id:344"

_Can be handy for internal use and for users that are not so comfortable with urls or that are on mobile._


Robiquet (commented):
I'm not really convinced on the use case for this as I think it's unusual that someone will know the market id (and the search syntax needed) but not anything about the title. Bit concerned that this could regress the current search functionality if markets have numbers in the title or description.

robhyrk:
If it affects the current search functionality negatively I'd say we should avoid implementing this. Personally I just put the id right into the URL. 

However, it got me thinking that we should include some sort of feature for users to be able to save a collection of markets.

PR #2195, opened 2024-01-18
Court enhancements

### LINKED ISSUE #2135: [Court] Display APR
We should display APR as it's a standard metric in DeFi for users to get an idea of returns. Probably we can replace one of these to make space:
image
Calculation for APR would be (inflation amount)/(total staked)

### LINKED ISSUE #2136: [Court] Show juror details
We have a component to display more details on jurors which is useful for users to select the one to delegate to

### LINKED ISSUE #2137: [Court] Show juror identities where possible

Robiquet:
closes #2135 
closes #2136
closes #2137


components/court/JurorsTable.tsx
```diff
@@ -32,44 +34,55 @@ const columns: TableColumn[] = [
     accessor: "status",
     type: "text",
   },
-  {
-    header: "",
-    accessor: "button",
-    type: "component",
-    width: "150px",
-  },
```
robhyrk, line 40: can this be removed?

lib/hooks/queries/useIdentities.ts
```diff
@@ -0,0 +1,75 @@
...
+  const [sdk, id] = useSdkv2();
+
+  const query = useQuery(
+    [id, identitiesRootKey, addresses],
+    async () => {
+      if (addresses && isRpcSdk(sdk)) {
+        const userIdentities: (UserIdentity | null)[] = [];
+        const res = await sdk.api.queryMulti(
+          addresses.map((address) => [
+            sdk.api.query.identity.identityOf as any,
```
robhyrk, line 18: is it possible to use a more defined type instead of `any` ?

robhyrk (approved):
looks good! just left a few suggestions

components/court/JurorsTable.tsx
Robiquet: Wanted to check with Jorn, we might keep it

lib/hooks/queries/useIdentities.ts
Robiquet: The api doesn't seem to have a type for this for some reason. I could write one, though the substrate types are a bit of a pain to write out manually and it wouldn't bring us much additional safety.

yornaath (approved):

PR #2201, opened 2024-01-19
Web 3 Wallet Update - Custom Modal

### LINKED ISSUE #2151: web3 auth wallet update
- [x] implement new web3 auth user flow and designs (designs needed)
- [x] migrate to custom modal version of web3 auth

robhyrk:
closes #2151 

note: will need to generate prod client keys for social logins


Robiquet (commented):
If I don't have a wallet installed I get shown this but on the next step it just points me in the direction on installing a wallet so I think it's a bit confusing. Think we need an additional branch to check if the user wants username/password signup or to download a wallet. 
image

package.json
```diff
@@ -27,6 +27,7 @@
     "@plaiceholder/next": "^2.5.0",
     "@polkadot/api": "^10.10.1",
     "@polkadot/extension-dapp": "^0.44.6",
+    "@polkadot/keyring": "^12.6.2",
```
Robiquet, line 30: Is this needed? I think we already inherit it from one of the other libraries

lib/hooks/useWeb3Wallet.ts
```diff
@@ -0,0 +1,173 @@
...
+        {
+          type: "Error",
+          autoRemove: true,
+          lifetime: 5,
+        },
+      );
+      return;
+    }
+    if (web3auth.connected) {
+      await web3auth.logout();
```
Robiquet, line 57: Might be a bit unexpected that a function that is called `login` also logs people out

package.json
robhyrk: are you referring to this: `import { decodeAddress } from "@polkadot/keyring";` ?

robhyrk:
> If I don't have a wallet installed I get shown this but on the next step it just points me in the direction on installing a wallet so I think it's a bit confusing. Think we need an additional branch to check if the user wants username/password signup or to download a wallet. image

I reverted to the original onboarding and will update that completely in the other onboarding PR I have open

lib/hooks/useWeb3Wallet.ts
robhyrk: this is just in case for when a user is already logged in but wants to switch to another provider without logging out. if this isn't there it'll throw an error stating the provider is already connected

package.json
Robiquet: Yeah this dependency

lib/state/util/web3auth-config.ts
```diff
@@ -1,9 +1,32 @@
...
+  [key: string]: {
+    verifier: string;
+    verifierSubIdentifier: string;
+    typeOfLogin: TypeOfLogin;
+    clientId: string;
+  };
+}
+
+export const clientId = isWSX
+  ? process.env.NEXT_PUBLIC_WEB3AUTH_CLIENT_ID_WSX
```
Robiquet, line 20: Why do we need 2 different variable names? Instead of just `NEXT_PUBLIC_WEB3AUTH_CLIENT_ID`

robhyrk: was using it switch between branches for wsx but it's not needed so I removed it

package.json
robhyrk: Do you know why this would cause a build error now on Vercel? I'm getting `Cannot find module '@polkadot/keyring' or its corresponding type declarations.` 

When I run `yarn cache clean --all` and `yarn install` I do notice that the `keyring` package inside `node_modules` gets removed when on this branch

Robiquet: Yeah that's strange I've just added it explicitly, should be ok for now

PR #2212, opened 2024-01-29
Update onboarding

robhyrk:
- [x] update onboarding prompts
- [x] confirm mobile steps
- [x] email notification? (created separate issue #2221)

NOTE: contains #2201 which should be merged first


components/account/WalletSelect.tsx
```diff
@@ -44,41 +44,73 @@ const WalletSelect = () => {
...
-                logoAlt={wallet.logo?.alt}
-                logoSrc={wallet.logo?.src}
-                hasError={hasError}
-                error={error}
-              />
-            );
-          })}
+        {isMobileDevice ? (
+          <a
+            href="https://novawallet.io/"
```
Robiquet, line 78: Could you open in new tab?

components/create/editor/Publishing.tsx
```diff
@@ -125,7 +125,7 @@ export const Publishing = ({ editor, creationParams }: PublishingProps) => {
           .mul(2)
           .plus(baseAssetTransactionFee ?? 0)
       : null;
-
+  console.log(foreignCurrencyCost);
```
Robiquet, line 128: Can be removed

pages/api/onboardUser.ts
```diff
@@ -0,0 +1,52 @@
+import { NextApiRequest, NextApiResponse } from "next";
+const SibApiV3Sdk = require("@getbrevo/brevo");
+
+export default async function onboardUser(
```
Robiquet, line 4: This feels like something that should be done server side as if anything goes wrong with this request the two lists are going to end up out of sync. 

If you want to go with this approach this endpoint also needs to check that the email is authenticated with web3auth otherwise anyone can use this endpoint to spam our mailing list with fake signups

Robiquet (approved):

PR #2228, opened 2024-02-05
Limit orders

Robiquet:


robhyrk:
Seems like an indexer related issue: when I cancel an order it remains there after cancelling the order and I get the following error: `orderbook.OrderDoesNotExist :: The order does not exist.`

robhyrk (approved):

PR #2246, opened 2024-02-08
Handle trusted markets.

yornaath:
- [ ] Filter markets with no dispute mechanism(trusted markets) created by non whitelisted accounts.
- [ ] Show "warning" on market pages for allowed trusted markets.


lib/hooks/queries/useInfiniteMarkets.ts
```diff
@@ -39,6 +39,10 @@ export type QueryMarketData = Market<IndexerContext> & {
   prediction: { name: string; price: number };
 };
 
+const WHITELISTED_TRUSTED_CREATORS = [
+  "dE2cVL9QAgh3MZEK3ZhPG5S2YSqZET8V1Qa36epaU4pQG4pd8",
```
Robiquet, line 43: What's this address? Could we use env variable for it?

yornaath: Done ✅

Robiquet (approved):

PR #2270, opened 2024-02-14
Protected IPFS api

yornaath:
Implement a protected IPFS api available to the app that validates metadata and its size.

**Note:**
When this is merged to mainnet app and goes live the cluster authentication username and password should be updated and updated in the env variables and then the app redeployed:


pages/api/ipfs/index.ts
```diff
@@ -0,0 +1,137 @@
+import { IOMarketMetadata } from "@zeitgeistpm/sdk";
+import type { PageConfig } from "next";
+import { NextRequest } from "next/server";
+
+const CLUSTER_ENDPOINT = process.env.IPFS_CLUSTER_URL;
+
+const headers = {
+  Authorization: `Basic ${Buffer.from(
+    `${process.env.IPFS_CLUSTER_USERNAME}:${process.env.IPFS_CLUSTER_PASSWORD}`,
```
sea212, line 9: Those will never be exposed to the end user?

pages/api/ipfs/index.ts
```diff
@@ -0,0 +1,137 @@
...
+  ).toString("base64")}`,
+};
+
+export const config: PageConfig = {
+  runtime: "edge",
+};
+
+export default async function handler(req: NextRequest) {
+  if (req.method === "POST") {
+    const [error, parsed] = IOMarketMetadata.validate(
```
sea212, line 19: It ensures that the form of the json matches what we expect as valid market metadata?

lib/metadata-storage.ts
```diff
@@ -0,0 +1,121 @@
...
+import { u8aConcat } from "@polkadot/util/u8a";
+import { MetadataStorage, createStorage } from "@zeitgeistpm/sdk";
+import { Codec, JsonCodec } from "@zeitgeistpm/utility/dist/codec";
+import * as O from "@zeitgeistpm/utility/dist/option";
+import * as Te from "@zeitgeistpm/utility/dist/taskeither";
+import { Storage, StorageError } from "@zeitgeistpm/web3.storage";
+import * as IPFSHTTPClient from "ipfs-http-client";
+import { CID } from "multiformats/cid";
+
+const node = IPFSHTTPClient.create({ url: "https://ipfs.zeitgeist.pm" });
```
sea212, line 11: Why is that hardcoded?

lib/metadata-storage.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export const createMetadataStorage = (): MetadataStorage => {
```
sea212, line 13: When is this expected to be used and when is the cluster expected to be used?

pages/api/ipfs/index.ts
yornaath: No those are only loaded on the server. Also vercel/next prevents any env varible that isnt prefixed with `NEXT_PUBLIC` from being sent to the client.

pages/api/ipfs/index.ts
yornaath: Yes

lib/metadata-storage.ts
yornaath: Doesnt need to be.

lib/metadata-storage.ts
yornaath: This is only used to read data from IPFS so this should be set to the new hosted ipfs node I believe. Cluster is used to add/pin files and remove them when rolling back a failed transaction.

lib/metadata-storage.ts
sea212: Hmm I think it is sufficient though, don't see why we would want to change this.

lib/metadata-storage.ts
sea212: We will shortly shut down the cluster and only use a singular hosted IPFS node (replication happens also in Subsquid). For now this is the right approach though, let's adjust once we have the hosted IPFS node ready.

sea212 (approved):

pages/api/ipfs/types.ts
```diff
@@ -0,0 +1,18 @@
+import * as z from "zod";
+
+export const IOMarketMetadata = z.object({
```
Robiquet, line 3: Might be nice to share with market form validation and add some rough size restrictions if possible

Robiquet (approved):

PR #2273, opened 2024-02-15
Trending market + latest Trades section 

Robiquet:


components/front-page/TrendingMarketsCompact.tsx
```diff
@@ -0,0 +1,82 @@
...
+  });
+
+  const { data: cmsMetadata } = useMarketCmsMetadata(marketId);
+
+  const prediction = getCurrentPrediction(market.assets, market);
+
+  return (
+    <Link
+      href={`/markets/${marketId}`}
+      className="flex h-[70px] items-center p-4 hover:bg-[#D4E7F4]"
```
robhyrk, line 48: is the background colour a one-off or might it be best to add it into tailwind config?

robhyrk (changes requested):
I noticed that latest trades are pulling from NTT markets. I think rather than excluding specific foreign assets in the queries filter we should filter to only include assets we use within Zeitgeist (DOT, ZTG, USDCD) in case we get more foreign assets like WSX/NTT in the future.

components/front-page/TrendingMarketsCompact.tsx
Robiquet: Not sure what the plan is, I just default to adding it like this until we get duplicates

Robiquet:
> I noticed that latest trades are pulling from NTT markets. I think rather than excluding specific foreign assets in the queries filter we should filter to only include assets we use within Zeitgeist (DOT, ZTG, USDCD) in case we get more foreign assets like WSX/NTT in the future.

Could you create a separate issue for this?

robhyrk (approved):

PR #2308, opened 2024-02-22
Prune some unused code + files

yornaath:
Pruning osme unused files and code using `knip`


lib/hooks/useLocalStorage.ts
```diff
@@ -1,33 +0,0 @@
-import { useState } from "react";
```
robhyrk, line 1: wondering if this might be good to keep for now? I see it was being used with the nfts

robhyrk (changes requested):
Seems all good to remove. Maybe @Robiquet should have a look too. 

Only thing I saw might be worth keeping was the local storage hook.

robhyrk:
Also I'm wondering if maybe the yarn.lock should be ignored? I'm not entirely sure how that might affect the build process

PR #2359, opened 2024-03-12
Claim page

Robiquet:


robhyrk (changes requested):
Looking good! 

One other thing I noticed was that I was able to make a claim twice with the same account and got a "Successfully claimed" message. Not sure if that's just because it's on testnet but wanted to bring it up just in case.

robhyrk (approved):

PR #2376, opened 2024-03-26
Twitch embed on market page

### LINKED ISSUE #2369: Display twitch stream on market page
Add twitch stream on market page. URL should be editable from the CMS

yornaath:
Added twitch embedded stream on market page.

- [x] Show stream tab if cms metadata has stream url or channelname.
- [x] Highlight stream tab button if channel is live.
- [x] Display stream tab by default if channel is live.
- [x] Added streamer url/name validation in the sanity cms.

Fixes #2369


Robiquet (changes requested):
Charts don't display anymore, for example: 
https://ui-git-tr-twitch-embed-zeitgeistpm.vercel.app/markets/809

Robiquet (approved):

PR #2383, opened 2024-04-05
New asset systems fixes + New indexer schema

Robiquet:


components/portfolio/CurrenciesTable.tsx
```diff
@@ -146,7 +145,13 @@ const CurrenciesTable = ({ address }: { address: string }) => {
         asset: (
           <ImageAndText
             name={balance.symbol}
-            imagePath={lookupAssetImagePath(balance.foreignAssetId) ?? ""}
+            imagePath={
```
robhyrk, line 148: Noticed that ZBS and DOT image is not showing up in my portfolio under Balances

robhyrk (changes requested):

PR #2438, opened 2024-06-04
Localize number formatting

### LINKED ISSUE #2059: Use of period/comma as decimal point is inconsistent across the app
Many fields use a decimal point, others a comma. I think we should go with a decimal point everywhere. I come from a region where it's customary to use a comma as a decimal point and a period as thousands separator, and I still would prefer using a period because that seems to me to be the standard in finance.

robhyrk:
closes #2059


Robiquet (commented):
Don't think this will close #2059 as it doesn't cover the formatting of number inputs

PR #2465, opened 2025-01-14
Misc Updates

robhyrk:
- remove airdrop btn and memecoin markets from front page

### DIFF
--- a/components/front-page/Topics.tsx
+++ b/components/front-page/Topics.tsx
@@ -21,11 +21,14 @@ export const Topics = ({
   return (
     <>
       <div className="hidden w-full gap-3 md:flex">
-        {topics.map((topic, index) => (
-          <Link
-            key={index}
-            href={`/topics/${topic.slug}`}
-            className={`
+        {topics.map((topic, index) => {
+          return (
+            //TODO: temp solution. waiting for login
+            topic.slug !== "memecoin-markets" && (
+              <Link
+                key={index}
+                href={`/topics/${topic.slug}`}
+                className={`
             ztg-transition flex flex-1 cursor-pointer items-center gap-4 rounded-lg p-3 transition-all md:max-w-sm md:hover:scale-[1.015]
               ${
                 (isString(selectedTopic) && selectedTopic === topic.slug) ||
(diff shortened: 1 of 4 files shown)


components/front-page/Topics.tsx
```diff
@@ -21,11 +21,14 @@ export const Topics = ({
...
     <>
       <div className="hidden w-full gap-3 md:flex">
-        {topics.map((topic, index) => (
-          <Link
-            key={index}
-            href={`/topics/${topic.slug}`}
-            className={`
+        {topics.map((topic, index) => {
+          return (
+            //TODO: temp solution. waiting for login
```
Chralt98, line 26: What do you mean by `waiting for login`? Do you need login access to the content management system, in which you can deactivate topics on the front end?

components/front-page/Topics.tsx
```diff
@@ -73,35 +80,39 @@ export const Topics = ({
...
-                <Link
-                  href={`/topics/${topic.slug}`}
-                  key={index}
-                  className="flex flex-1 cursor-pointer items-center gap-4 rounded-lg bg-white p-2 transition-all hover:bg-gray-200 hover:bg-opacity-30 md:max-w-sm"
-                >
-                  <div className="relative h-10 w-10">
-                    <Image
+              {topics.map((topic, index) => {
+                return (
+                  topic.slug !== "gaza" && (
```
Chralt98, line 85: I think you meant `memecoin-markets` here instead of `gaza`, but anyways... I don't think it's optimal to exclude the `memecoin-markets` here in this `Topics` component, because [this line](https://github.com/zeitgeistpm/ui/blob/7aae53bbda6edd0c35fc4e8d3a31d462be170ed3/pages/index.tsx#L176) indicates that it will select the first entry from the CMS, which is the memecoin-markets entry. I would rather wait for the access of the CMS instead of doing the dirty fix like here. When will you get access to the CMS? If it took too long to gain access to the CMS, I would suggest changing [this line](https://github.com/zeitgeistpm/ui/blob/7aae53bbda6edd0c35fc4e8d3a31d462be170ed3/pages/index.tsx#L176) to skip to the next item from the CMS if the `memecoin-markets` topic is going to be selected.

components/top-bar/index.tsx
```diff
@@ -219,11 +219,7 @@ const TopBar = () => {
         </div>
         <MarketSearch />
         <div className="center relative ml-auto gap-3">
-          {process.env.NEXT_PUBLIC_SHOW_AIRDROP !== "true" ? (
-            <GetTokensButton />
-          ) : (
-            <AirdropButton />
-          )}
+          <GetTokensButton />
```
Chralt98, line 222: Looks good. You could think about removing all related components (including claim.tsx) for the Airdrop button, since we wouldn't use it anymore.

Chralt98:
E2E Tests failing in the CI?

components/top-bar/index.tsx
robhyrk: In case we decide to do another airdrop for whatever reason I'll leave it  for now

components/front-page/Topics.tsx
robhyrk: yes, thank you. haven't touched Zeitgeist code for a while so just getting back to it. I saw your comment from earlier on DIscord and have removed the sections as per your suggestion

robhyrk:
> E2E Tests failing in the CI?

 I'm not sure if these are even setup correctly so I'll need to look into it. Might just disable until we can get a fix on it

Chralt98 (approved):

PR #2470, opened 2025-02-02
Banxa Disclaimer

### LINKED ISSUE #2469: Banxa Disclaimer
⁠Prior to redirecting the customers to the checkout page, please add a disclaimer that they will be redirected to Banxa, a third-party service provider. This is regardless if you are going to use our checkout page or use your own. I will be sending some samples for reference.

robhyrk:
closes #2469


saboonikhil (approved):
Thank you!

